PR #56, opened 2014-12-17
Simplify package dependency installation

djvine:
This is just a suggestion.

Packages could be installed simultaneously with
```python
pip install -r requirements.txt
```
So my pull request shows an implementation of this. I like it because it it more automated but the downside is that dependencies now live in two places: (i) setup.py install_requires and (ii) requirements.txt.

Anyway - just a thought!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,0 +1,5 @@
+numpy>=1.8.0
+scipy>=0.13.2
+h5py>=2.2.1
+pywavelets>=0.2.2
+scikit-image>=0.9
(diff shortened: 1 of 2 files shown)


dgursoy:
@djvine's option seems a nice option considering easier (independent) management of virtual environments:

One can simply use:
```pip freeze > requirements.txt```
```pip install -r requirements.txt```

as described here:
http://docs.python-guide.org/en/latest/dev/virtualenvs/

PR #66, opened 2015-01-29
X-ray Fluorescence Tomography

djvine:
This pull will add the capability for tomopy to handle X-ray fluorescence datasets. The major change being that XFT datasets are 4-dimensional. The 4-th dimension is the channel or element.

Tomopy has been modified to add a new directory under "tomopy" called xftomo (similar to xtomo). The structure exactly mimics (and steals code from) xtomo.

The main addition to the algorithms/preprocess folder is the align_projections function which uses a few different methods to align XFT projections before tomo reconstruction.

A demo was added demo/demo_xtomo.py demonstrating usage.

There is lots to do with documentation.

Feedback/criticism is welcomed.


tacaswell:
attn @licode

.gitignore
```diff
@@ -1,60 +1,41 @@
-# Compiled source #
```
djvine: These changes to the .gitignore weren't intentional. Looks like I somehow overwrite it with the one I usually use.

newville:
@djvine: Thanks - I think this will be very useful!   Could you give an overview of what the data format and layout is assumed to be?   It looks like I should be able to write something like `xftomo_io.import_aps_13ide()`, but I'm not 100% sure what the output should be.  It looks like it goes like

```python
  data, theta, names = import_aps_XXX()
```
where `data` is (Energy_index, Theta_index, Y_index, X_index), `theta` is the array of angles, and `names` is a string for Energy names (as if ROI data?).  Is that right?   Am I missing anything?

Eventually, we'd like to be able to pass in the full Energy array  (but have only 1 or 2 slice so data-size is not 1000^4, but only 1000^3) and be able to reconstruct to the fluorescence spectra per voxel, rather than just get an image for each pre-defined ROI.  Doga has done examples of this with some of our data, but we'd love to be able to do this on a regular basis.

Also: would it be sensible to have value arrays for all the dimensions?   If Theta is required, why not also require the values for X, Y, and Energy?

Anyway, thanks!!

djvine:
@newville The data format are assumed to be [channel, projection, slice, pixel] or equivalently [energy, theta, y, x]. For 2ide the "channel" dimension is usually the quantified elemental concentration from MAPS. However, for the energy index if you want to use the full 2048 energy spectra it should all work fine - each channel is reconstructed separately. 

The only required variables are `data` and `theta`. `names` are optional and default to just channel number if they're not given. I just used this for naming tif when saving.

A single slice of the full spectra could still be accommodated like [energy, theta, 1, pixel]. The projection alignment methods using cross-correlation won't work but I just added a new method yesterday which fits the sinogram to a sine curve which should work.

I can see why requiring all the value arrays makes sense. For example if the arrays are up- or down-sampled during the reconstruction or to implement the kind of reconstruction Doga has demonstrated where neighbouring energy channels influence each other. So I think that's probably a good idea. There are so many more features to add I just didn't get that far yet!

rh66:
David,
I am not sure my “replying directly” is the right way of sending email to you, but hope you can get this email.
May I confirm with you that the fluorescence tomopy does not include any fluorescence absorption correction, right?
I don’t know much about tomography, but wondering when you say “projection alignment methods … fits the sinogram to a sine curve”, do you have the reference included in your code header or comment field, or somewhere the document is?
Thanks,
Rong
-------------------
Rong Huang
Cornell High Energy Synchrotron Source
Cornell University
Ithaca, NY  14853
USA
Phone:  1-607-2553753
Fax:        1-607-2559001

djvine:
Hi Rong,

there is no absorption correction yet. As of right now all the new software
does is sequentially apply the existing tomopy reconstruction techniques
(mlem, pml...) to each channel of the fluorescence tomo dataset. A self
absorption correction would be ideal and we have some plans at APS to
implement it eventually.

I implemented a few projections alignment methods which are documented a
little in the tomopy/algorithms/preprocess/align_projections.py. To be
honest the cross correlation methods are much more robust but I tested the
sinogram curve fitting on synthetic data and it should work. The key is
choosing a sinogram slice with a strong "sine" feature - if that makes
sense - for the least squares fitting to work. At the bottom of
align_projections.py under "if __name__ == '__main__'" i have a test
implementation so you can see how it should work.

I'm happy to answer any questions. If you see a place for improvement
please feel free to fork the repo and make a pull request!

Best,
David

David

PR #69, opened 2015-02-13
Fixed demo_xftomo.py to only include simulated data.

djvine:
Removed references in demo_xftomo to real data. It now only uses simulated data. I chose not to include real data because we still don't have a reliable alignment method for the projections so the reconstructions of real data don't look so good. The simulations work great though!


dgursoy:
Thanks David!

PR #72, opened 2015-03-23
Do not call plt.style if not available

matze:
Ubuntu 14.04 is shipped with Matplotlib 1.3.1 which does not include this API. On the other hand it might also not be a very good idea to override the user's decision how things should look like.


dgursoy:
Yes, correct. These lines can even be removed.

matze:
Do you intend to do it or should I open a new pull request?

dgursoy:
I think we can leave it for now. Thanks!

PR #84, opened 2015-05-01
Merging updates from TomoPy master.

carterbox:


dgursoy:
The old tomopy is now at legacy branch. You can submit a PR to there if the patch is about the old one.

carterbox:
I was actually trying to pull updates into my branch from tomopy/tomopy. Apparently you can't delete pull requests, you can only close them.

PR #89, opened 2015-05-18
TomoPy 0.1.9

dgursoy:
Improvements
-------------------
- Bibtex files are put in a separate folder

Bug fixes
------------
- Centering corrected


meta.yaml
```diff
@@ -1,6 +1,6 @@
 package:
   name: tomopy
-  version: '0.1.8'
+  version: '0.1.9'
 
 source:
   git_url: https://github.com/dgursoy/tomopy.git
```
tacaswell: If you are targeting a git repo for a fixed release you should also include a `git_rev` entry other wise if you re-run old recipes you will get the new code version with the old number.

dgursoy: I haven't thought about using old versions for now. How about using ``git_tag`` instead of ``git_rev``? Seems more cleaner.

tacaswell: `git_tag`, `git_rev` and `git_branch` are all aliases of each other for now.

Another useful trick is to use a templated `meta.yaml` (see https://github.com/Nikea/scikit-xray/blob/master/conda-recipe/meta.yaml) which will set the version string based on the current commit and most-recent tag (ex v0.1.9.post7_g012a456 ) so that conda builds off of any point in master will sort correctly.

dgursoy: I like the idea but shouldn't it be better if it is the same as the one described in setup.py?

tacaswell: There are 3 places you need to keep track of the version: meta.yaml,
setup.py, and __version__ in the top level `__init__.py`.  Of the three I
think getting the conda one right is most important (so that your
non-release builds sort right and so you can always get back _exactly_ what
a given version was), then the `__version__` as that is what users tend to
see and last the version in `setup.py` as it is only used by pip and I only
marginally care about supporting that.

The way I normally deal with that is to between releases change the version
in `setup.py` and `__init__.py` to `vX.Y.Z.post0`  so it indicates that it
isn't an 'official' release but is some intermediate after `vX.Y.Z` and let
the conda packages keep the real version.  There is a tool (versioneer)
which will handle all of this automatically, @danielballan has experience
with it, but I have not had the bandwidth to get it set up the way I want.

dgursoy: Thanks Tom for the detailed information. i have few more points to ask but we can discuss more on this in few days.

PR #90, opened 2015-05-19
Fix negative values in shepp3d phantom.

francisco-dlp:
I expect the shepp3d phantom to be > 0, however small negative values are present:

```python

>>> tomopy.io.phantom.shepp3d((6, 8, 10)).min()
-1.4901161e-08

```

This hopefully fixes it.


dgursoy:
It was the float operations that were causing the problem. Thank you for fixing this.

francisco-dlp:
Thank you for this nice package.

PR #96, opened 2015-06-15
Add ASTRA wrapper for reconstruction

dmpelt:
This PR adds a wrapper for the ASTRA toolbox for reconstruction. It uses the calling method introduced in #95. Extra options are:

    method : str
        ASTRA reconstruction method to use.
    num_iter : int, optional
        Number of algorithm iterations performed.
    proj_type : str, optional
        ASTRA projector type to use: 'line', 'linear', or 'strip' (CPU only).
    extra_options : dict, optional
        Extra options for the ASTRA config (i.e. those in cfg['option'])

One thing I found out is that CUDA does not combine well with the `multiprocessing` module. So when using cuda, only a single processor can be used (which is not a big problem, the parallelization is done by the GPU). Multi-GPU support is currently not supported, but it might be able to add this later.

Example reconstruction:

    >>> import tomopy
    >>> obj = tomopy.shepp3d() # Generate an object.
    >>> ang = tomopy.angles(180) # Generate uniformly spaced tilt angles.
    >>> sim = tomopy.project(obj, ang) # Calculate projections.
    >>>
    >>> # Reconstruct object:
    >>> rec = tomopy.recon(sim, ang, algorithm=tomopy.astra,
    >>>       options={'method':'SART', 'num_iter':10*180, 'proj_type':'linear',
    >>>       'extra_options':{'MinConstraint':0}})
    >>>
    >>> # Show 64th slice of the reconstructed object.
    >>> import pylab
    >>> pylab.imshow(rec[64], cmap='gray')
    >>> pylab.show()

### DIFF
--- a/tomopy/__init__.py
+++ b/tomopy/__init__.py
@@ -62,6 +62,7 @@ from tomopy.misc.morph import *
 from tomopy.prep.normalize import *
 from tomopy.prep.phase import *
 from tomopy.prep.stripe import *
+from tomopy.recon.wrappers import *
 from tomopy.recon.algorithm import *
 from tomopy.recon.rotation import *
 from tomopy.sim.project import *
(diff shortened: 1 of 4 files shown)


dmpelt:
Added ASTRA example to reconstruction documentation.

PR #105, opened 2015-07-23
Doc/demo updates

decarlof:
It seems that .automodule needs a "def function" in the python script demo to create the link to the source code in the doc page.


decarlof:
Initial set of APS reconstruct examples is completed; will start slowly adding examples for other facilities

PR #117, opened 2015-08-18
print statement -> print function

ericdill:


tacaswell:
You probably need `from __future__ import print_function` on all of these.  I think this will error on python2 as-is.

dgursoy:
Fixed in #118

PR #118, opened 2015-08-18
print function fix

dgursoy:


tacaswell:
The future import must be the first import, this will raise a `SyntaxError`

dgursoy:
Thanks (and I need coffee!)

PR #123, opened 2015-09-02
Added example for Petra III

decarlof:
I received an email from P05 user asking how to use tomopy on their data. Here is the answer.
Tested on Petra III data bank data set.

Added run requirement to use dxfile (data exchange writer using David's API).

Corrected minor typo.


dgursoy:
Thanks Francesco.

PR #126, opened 2015-10-13
ncore now correctly defaults to cpu_count instead of 1.

michael-sutherland:
When ncore defaults to None, ncore <= 0 evaluates to True.  This results in ncore being set to 1 by default.  I moved the <= 0 check to a point after None values being replaced with the number of cpus to fix it.


dgursoy:
Thanks @michael-sutherland. The indentation seems to be not correct though. Can you correct this?

michael-sutherland:
Good catch!  The tabs should now be replaced with spaces.

dgursoy:
Good to go! Thanks.

PR #128, opened 2015-10-15
Windows fix 1

aglowacki:
These changes add the defines for windows dll to export the c functions.


dgursoy:
Thanks! Does this also help x64 systems?

Issue #127, opened 2015-10-15
Gridrec segfault when running with specific number of cores


dmpelt:
Using both the latest conda build of tomopy (0.1.11) and the latest git master, I get a segfault when running `gridrec` with a specific number of cores. The result is the problem of #120, where you have to kill the parent python process in order to regain control of the terminal. The following script shows the problem on my machines:

    import tomopy
    import numpy as np
    ang = tomopy.angles(256)
    prj = np.zeros((256,100,256),dtype=np.float32)
    tomopy.recon(prj,ang,algorithm='gridrec',ncore=1) # works
    tomopy.recon(prj,ang,algorithm='gridrec',ncore=2) # works
    tomopy.recon(prj,ang,algorithm='gridrec',ncore=5) # works
    tomopy.recon(prj,ang,algorithm='gridrec',ncore=3) # works
    tomopy.recon(prj,ang,algorithm='gridrec',ncore=8) # hangs

The following `gdb` log shows the location of the segfault:

    #0  0x00007f7198ed1435 in gridrec (data=0x7f71d71de000, dx=256, dy=100, dz=256, center=0x7f7199213aa4, theta=0x0, recon=0x7f71910cf000, ngridx=256, ngridy=256, fname=0x1b62eb0 "shepp", istart=91, iend=100)
    at src/gridrec.c:311

Maybe something goes wrong earlier, since `theta` is already `0x0` in the function call. 

Some other combinations of numbers of angles, slices, detectors, and cores segfault as well, but many others work fine. I did not find a clear pattern yet...

dgursoy:
This more seems to be a bug in python multiprocessing, though not yet clear. @aglowacki is investigating this and may have some better insights about the possible source.

dmpelt:
I tried removing the multiprocessing parts to test it out, but the segfault kept appearing. 

In the end, I may have found a possible cause for this particular segfault, and why it only happens for specific number of cores and slices! It turns out that the segfault only happens when the final chunk starts with an odd slice index.

I submitted a PR with a fix in #129.

dgursoy:
I assume this is now fixed.

PR #129, opened 2015-10-19
Fix segfault in gridrec for odd starting slice in final chunk

dmpelt:
When the final chunk of a gridrec reconstruction starts with an odd slice number, `islc2` will point to an invalid memory address (one slice after the final one) in the last loop iteration. A previous check was already there to prevent these cases, but it only checked for when the starting slice was equal to the final slice. By changing this to check for the current slice that is reconstructed, I think the bug is fixed. 

I do not really know how the gridrec implementation is supposed to work, so please check whether I break something by this change!

PS. This is what caused the segfault and hang in #127 (since 100 slices over 8 cores will make the final chunk start at slice 91). After the fix, I don't get a segfault any more.

### DIFF
--- a/src/gridrec.c
+++ b/src/gridrec.c
@@ -170,7 +170,7 @@ gridrec(
             while(j<dz)  
             {     
                 sino[j].r = data[j+s*dz+p*dy*dz];
-                if (!(dy == 1 || iend-istart == 1))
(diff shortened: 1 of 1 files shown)


dgursoy:
Thanks! Good catch!

Issue #120, opened 2015-08-27
Alternate reconstruction methods do not complete.


carterbox:
Ubuntu 15.04 / conda tomopy 0.1.11 / AMD processor

I was trying to test different reconstruction methods to compare the results:

```
for algo in ['gridrec', 'mlem', 'osem', 'pml_hybrid', 'ospml_hybrid']:
        rec = tomopy.recon(proj, theta, center=1274.2, algorithm=algo, emission=False)
        print algo + " reconstruction complete"

        tomopy.io.writer.write_tiff_stack(rec, fname='./recon_test/' + algo)
        print algo + " SAVED!"
```

Here is the output.

```
carterbox@COMPY:$ python new_tomopy_script.py
phase retreval complete
gridrec reconstruction complete
gridrec SAVED!
mlem reconstruction complete
mlem SAVED!
```

Everything works as expected until the workers all finish doing their parts of the 'osem' reconstruction. After their processors usage drops to zero, the workers are never deallocated. They just sit there indefinitely with CPU usage at zero. The same happens for both of the above hybrid methods as well.

When I try and interrupt the process using ^C it looks like all the workers are waiting for more tasks. Here is an abbreviated list of the traceback from the interrupt. Also, after using ^C to interrupt the process, new workers are created to replace the ones that were killed.

```
^CProcess PoolWorker-56:
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/process.py", line 114, in run self.run()
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/process.py", line 114, in run self._target(*self._args, **self._kwargs)
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/queues.py", line 378, in get
    self.run()
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/pool.py", line 102, in worker
    task = get()
  File "/usr/local/anaconda/lib/python2.7/multiprocessing/queues.py", line 376, in get
    racquire()
```

Is there a debug mode for tomopy? Since I updated from the legacy version, there seems to be no terminal output as things progress.

dakefeng:
can you try taking out the setting for center?

carterbox:
Removing the center specification doesn't work.

tomopy:
[...]

PR #130, opened 2015-10-20
Fix gridx/gridy producing out of bounds indices

dmpelt:
This solves a segfault I was getting with `sirt`, but it may solve other segfaults for iterative methods as well (maybe the original #120 error).

After some debugging, I found out that in the `sirt` method, some pixel indices were becoming larger than their allowed values (which caused them to read/write outside of array bounds later).

It turns out that in the `trim_coords` method, `coordx`/`coordy` values outside of the reconstruction grid are checked for in the `if` statements,  but `gridx`/`gridy` values could still be outside (actually, on the exact edge) of the reconstruction grid, which causes the problems.

By changing the bounds of the `for` loop, the problem is fixed, since only `gridx[0]`, `gridx[ry]`, `gridy[0]`, and `gridy[rz]` cause problems.

As in #129, I don't know whether this change can cause any problems at other points in the code, since I am not really familiar with the code, so please check!

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -134,7 +134,7 @@ trim_coords(
 
     *asize = 0;
     *bsize = 0;
-    for (n=0; n<=rz; n++) 
+    for (n=1; n<rz; n++) 
     {
         if (coordx[n] > gridx[0]) 
         {
@@ -146,7 +146,7 @@ trim_coords(
             }
         }
     }
-    for (n=0; n<=ry; n++) 
+    for (n=1; n<ry; n++) 
     {
         if (coordy[n] > gridy[0]) 
         {


dgursoy:
I am confused a little bit, because ```grid[0]``` refers to the edge of the pixel but not the pixel itself (number of edges should be one larger than the number of pixels) . So by removing ```grid[0]``` and ```grid[rx]```you are basically removing the outer frame of the reconstruction (at least the way I see). Are you sure if the final reconstructed image boundaries are updated after this change in each iteration (or are they simply not updated and stay as ```1e-6```)?

dmpelt:
Indeed, I am not really sure about this myself as well. I will check what happens to the boundaries. The travis build errors seem to indicate that the boundaries indeed stay at `1e-6`. However, without this fix I get segfaults when using `sirt`, so I don't know what the correct solution is. The segfault occurs in the last slice, since it will read/write beyond the array bounds, but in other slices, it will incorrectly read/write values of adjacent slices.

What I noticed is that `gridx[ry]` is equal to `ry/2` (when the center of rotation is in the exact center), 
which means that later, it will result in the use of an index of `ry`. Since the reconstruction grid only has `ry` rows/columns, this index is out of bounds, which leads to a segfault. It could be that an added  `+/- 0.5` and/or some rounding operation somewhere can fix things, but I am not sure.

dgursoy:
Aha.. I then would suspect on one of the four corner pixels. Could you please give it a quick try after your fix, if one of the corner values (or any of the the outer frame values) is updated accidentally or not?

dmpelt:
I just checked, and after the fix the corner values are the only ones that are _not_ updated (so they are still at `1e-6`). The other values on the edges are updated. Does that make sense? I am not really sure what the correct behavior should be...

dgursoy:
After your fix or before your fix?

dmpelt:
After my fix. Before my fix, all values are updated (I just checked).

dgursoy:
This is good then!  I believe the real source of the problem lies in updating the index value of the pixels (i.e., line 234 of ```calc_dist``` function in ```utils.c```). Probably ```i1``` and ```i2``` are calculated incorrectly for the boundary frame depending on odd or even number of grid size.

dmpelt:
Indeed, I already thought that there might be a better solution than the one in this PR!

I think what happens is that in `calc_dist`, the `i1` and `i2` variables will end up being in the range of `0` to `ry`, including `ry` (when `x1` or `x2` is exactly equal to the extreme edge), which is out of bounds. Do you know how to fix this?

By the way, I believe that the `(i1>x1)` and `(i2>x2)` calls in `calc_dist` are always `false`, since `i1` is equal to `floor(x1)`. Is this related to the problem?

dgursoy:
I have to look into the code and make some debugging before I can comment to this one..

dgursoy:
I finally spent some time to go over the code. It seems that sometimes the indexing for the data may fall outside the grid because of imprecise float operations (i.e. ```trim_coords``` fail sometimes). In the new version I just made sure that this doesn't happen. I am closing this one and we can discuss on a new issue if this doesn't resolve the problem.

PR #131, opened 2015-10-20
Output files - dtype uint16

FedeMPouzols:
In `dtype.py` `as_uint16()` seems to return an int32 array.
https://github.com/tomopy/tomopy/blob/master/tomopy/util/dtype.py#L101-L103

Should `return as_dtype(arr, np.int32)` be `return as_dtype(arr, np.uint16)`?


dgursoy:
Exactly. Thanks!

Issue #133, opened 2015-11-07
Simple typo in pad function?


lbluque:
The function reads like this...
```python
def pad(arr, axis, npad=None, val=0):
    if npad is None:
        npad = _get_npad(arr.shape[axis])
    pad_width = _get_pad_sequence(arr.shape, axis, npad)
    cval = _get_pad_sequence(arr.shape, axis, npad)
    return np.pad(arr, pad_width, 'constant', constant_values=cval)
```
I believe that 
```python
cval = _get_pad_sequence(arr.shape, axis, npad)
```
should simply be
```python
cval = val
```
or remove that variable altogether:
```python
def pad(arr, axis, npad=None, val=0):
    if npad is None:
        npad = _get_npad(arr.shape[axis])
    pad_width = _get_pad_sequence(arr.shape, axis, npad)
    return np.pad(arr, pad_width, 'constant', constant_values=val)
```

dgursoy:
Correct. This made me realize few more points to correct. E.g. ```npad``` should be changed with sth like ```npad_left``` and ```npad_right``` (line 109), since if ```arr``` size is odd, then we can't have a power of 2 with fixed padding for both sides. Thanks for bringing that up.

PR #134, opened 2015-11-09
Update morph.py

dgursoy:
Changes:
- #133 fix
- Nghia's approach for automatically finding rotation center (Needs to be tested)


dgursoy:
Need tests and some refactoring though...

PR #144, opened 2015-12-09
ALS 8.3.2 hdf5 compatibility and new normalization and find center functions.

lbluque:


lbluque:
The `normalize_nf` and `find_center_pc` do not have unit tests yet. Need to work on those.

PR #149, opened 2015-12-11
remove_ring function and extended pad function to allow edge values

lbluque:


dgursoy:
Thanks a lot @lbluque!!

VERSION
```diff
@@ -1 +1 @@
-0.1.12
+0.1.14
```
lbluque: @dgursoy Not sure if this change was overlooked. Please make sure this is the version number tomopy should currently be on.

dgursoy: @lbluque correct!

PR #159, opened 2016-02-29
Add multi-GPU support for ASTRA

dmpelt:
This PR adds multi-GPU support for the ASTRA toolbox, splitting the slices over the GPUs specified by `gpu_list`. 

For Python 2, this means that the `futures` package is an optional dependency (only needed when using multi-GPU), but I didn't include it in the conda dependency list since I don't expect many people will use both multi-GPU and Python 2, and if it is not installed the error message (an import error for `concurrent.futures`) is clear enough to find a solution quickly.


dmpelt:
I added a bit more information about the projector types. More information can be found in de ASTRA documentation, specifically: http://www.astra-toolbox.com/en/latest/docs/proj2d.html.

dgursoy:
Awesome thanks. Nice docs btw!

PR #160, opened 2016-02-29
Catch exceptions and segfaults in child processes

dmpelt:
This PR tries to improve user-friendliness a bit when an error or exception occurs when running things with multiprocessing. When an error occurs in a child process, tomopy will not completely hang after this PR, which 'solves' #120. Of course, the underlying problem that caused the segfault/error won't be solved, but at least a user can keep working without needing to kill the main process. This PR will also catch keyboardinterrupts, so you can use ctrl+c to stop an execution.

I tried to use `concurrent.futures` first, with a `ProcessPoolExecutor`, but that did not support an initializer function, so I changed it to the implementation suggested [here](http://stackoverflow.com/questions/24396147/retrieve-exit-code-of-processes-launched-with-multiprocessing-pool-map/24396655#24396655).

Right now it polls the child processes each second, and if there is an error, it will take the whole pool down. I did not test the impact on performance of this, if it has a high impact we might want to increase the timeout.

### DIFF
--- a/tomopy/util/mproc.py
+++ b/tomopy/util/mproc.py
@@ -163,9 +163,26 @@ def _start_proc(arr, args):
         mp.Pool(processes=len(args),
                 initializer=init_shared,
                 initargs=(shared_arr, queue))) as p:
-        p.map_async(_arg_parser, args)
-    p.close()
-    p.join()
+        if p._pool:
+            proclist = p._pool[:]
+            res = p.map_async(_arg_parser, args)
+            try:
+                while not res.ready():
+                    if any(proc.exitcode for proc in proclist):
+                        p.terminate()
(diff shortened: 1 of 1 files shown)


xianghuix:
May the current solution cause memory leakage? Will a new pool be automatically created or the hanging job be killed, after the whole pool is taken off?

dgursoy:
Thanks @dmpelt!! Let's give it a try!

dmpelt:
@xianghuix : I don't expect this to cause memory leakage. The child processes will all be terminated, so won't exist after an error and can't leak memory, and the host process proceeds normally, so it can garbage collect the used memory. I also expect the `pool.terminate()` call to clean up all the shared memory. About the creation of a new pool: the host process will raise an error and will stop running, so no new pool is created.

PR #162, opened 2016-03-01
Reorganized mproc and sped up gridrec using FFTW

michael-sutherland:
There are a lot of changes.  Let me know if there are and issues and I'll try to fix them before we merge.  I know one potential issue is that the defaults for project and recon don't match (project is in sinogram order and recon is in radiograph order).  We can change the default for either.  Suggestions?


dgursoy:
Once the conflicts (I think because of the recent PR #162) are solved I'd very much like to accept this PR. Please don't forget to increase the version to 0.2.0 since the backward compatibility is slightly broken.

dgursoy:
```project``` and ```recon``` match issue should be easily resolved I believe. Any chance you can fix that one before the merge?

**Comment by michael-sutherland** [comment]
Any preference on the project/recon matching issue?  My plan is to add the sinogram_order parameter to project as well.  Should the default be sinogram order or radiograph order?

**Comment by michael-sutherland** [comment]
I've fixed the tests I could and think I've got everything updated.  Could someone else check the last failing tests?

**Comment by michael-sutherland** [comment]
The -lm links to the math library.  I think I needed it for the complex data type being used in gridrec, although I don't remember for sure.

**Comment by prjemian** [comment]
The -lm compiler option is necessary for almost any C/C++ code involving 
math.

**Comment by lbluque** [comment]
Travis CL build is failing. Does this have to do with fftw dependency?

File: meta.yaml [hunk 1]
```diff
@@ -1,6 +1,6 @@
 package:
   name: tomopy
-  version: '0.1.15'
+  version: '1.0.0'
```
lbluque: Should this be changed back to 0.2.0?

dgursoy: Just realized we set this up in docs, so I increased first one: The minor version is incremented for releases which add new, but backward-compatible, API features, and the major version is incremented for API changes which are not backward-compatible.

PR #163, opened 2016-03-02
Use correct python version in travis builds

dmpelt:
When working on travis-ci integration for ASTRA (for which I learned a lot from TomoPy's integration, thanks!), I noticed that TomoPy's travis configuration always uses python 2.7, even when testing in Python 3 mode. The reason for this is that the current script always uses `Miniconda-latest`, which is Python 2.7, and does not specify `python=$TRAVIS_PYTHON_VERSION` to conda to change the python version.

This PR updates the travis configuration to use the specified python version in conda. I also had to remove Python 2.6 from the test matrix, since conda does not seem to work with that version anymore.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,17 +12,21 @@ os:
 sudo: false
 
 before_install:
-  - wget http://repo.continuum.io/miniconda/Miniconda-latest-Linux-x86_64.sh -O miniconda.sh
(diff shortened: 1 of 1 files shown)


dgursoy:
Aha. It made me think why they don't use miniconda2 for the distribution. Thanks for noticing and correcting this!

PR #164, opened 2016-03-03
Fits IO: added reader

yxqd:
Doga, please check and see if you are OK with these additions. Once this pull request passed, I will try to add another for Fits writer.


tacaswell:
The astropy dependency should be noted in the documentation (someplace).

It is probably also worth skipping the tests if astropy is not installed.

decarlof:
Just a curiosity: who is using fits for tomograhy?

yxqd:
Fits data format is currently in use at CG-1D, a neutron imaging beam line at HFIR of ORNL

decarlof:
ok thanks

dgursoy:
Is there a more light-weight reader for fits than astropy? 

How about we make it optional like ```astropy = _check_import('astropy')``` like any other optional formats in line 74-76? Also we should have it in the condo-recipe: ```meta.yaml```.

decarlof:
@dgursoy there was a PyFITS but it now merged into astropy.  FITS is just a binary with an asci header. The header tag definition is for astronomical object so is this case they are not effectively storing any metadata except bit and data type and they are reading these as ascii string (stream). The FITS data starts after the tag "END". To load it they use astropy but given the very simple structure I am confident this can be replaced by a binary read => no need of astropy.

tacaswell:
Picking up an optional heavy dependency (I would not add it to the conda recipe) seems better than starting down the road of rewriting said dependency.

dgursoy:
@yxqd Is this still active? Any updates on this PR?

yxqd:
Hi Doga,

Thanks for the reminder. I was away last week and will try to get back to
this soon.

--Jiao

Dr. Jiao Lin
Computational Scientist
Oak Ridge National Lab

dgursoy:
Thanks!

dgursoy:
I'll close this since io will be handled by dxchange after the #182. The plan is to decouple data io and tomopy. Please follow it at: http:/dxchange.rtfd.org

decarlof:
@yxqd Hi Jiao
I was wondering if I can have  a full data set I can use as a template to add to my DXchange validation tests. If so please so please share your globus id and I will give you write permission at:

https://www.globus.org/app/transfer?origin_id=5e6eeff0-00c4-11e6-a71b-22000bf2d559&origin_path=%2F

thanks

francesco

yxqd:
Sorry I just realized that I missed your message earlier. My globus id is "linjiao".

decarlof:
@yxqd ok I eneabled you globus id fro friting. Please drop the dataset in the linjiao folder

yxqd:
Thanks. Uploaded.

decarlof:
@yxqd thanks. Can you give me the full name and a short description of the facility? I will include it in the [DXChange](http://dxchange.readthedocs.io/en/latest/source/demo.html).

yxqd:
* Name: HFIR CG-1D at ORNL
* Description: The CG-1D Cold Neutron Imaging Facility

yxqd:
@decarlof Hi Francesco, I wonder if the fits engine in dxchange is ready to use? Thanks!

decarlof:
@yxqd Hi Jiao the reader works but in order to load the data set and do a reconstruction I have some issue generalizing the naming  convention indexing. The importer is expecting  a base_name_XX.fits where XX is the projection index. Your naming convention has a base_name containing other indexing that prevents the data sorting at load time.

decarlof:
@yxqd Jiao can you share the code you currently use to load you projections?

PR #167, opened 2016-03-09
Develop

lbluque:
Fixed a counting error in normalize_nf and fixed unicode_literals error caused by python 3 support

### DIFF
--- a/tomopy/misc/morph.py
+++ b/tomopy/misc/morph.py
@@ -73,7 +73,7 @@ __all__ = ['pad',
 LIB_TOMOPY = extern.c_shared_lib('libtomopy')
 
 
-def pad(arr, axis, npad=None, mode='constant', **kwargs):
+def pad(arr, axis, npad=None, mode=str('constant'), **kwargs):
     """
     Pad an array along specified axis.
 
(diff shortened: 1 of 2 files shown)


lbluque:
Yes, completely unnecessary. removed.

tomopy/misc/morph.py
```diff
@@ -73,7 +73,7 @@
 LIB_TOMOPY = extern.c_shared_lib('libtomopy')
 
 
-def pad(arr, axis, npad=None, mode='constant', **kwargs):
+def pad(arr, axis, npad=None, mode=str('constant'), **kwargs):
```
tacaswell: Why add the `str` here?  There should be no reason for that.

tacaswell:
Ln 108 should probably use `six.text_type`

tomopy/misc/morph.py
lbluque: So the unicode_literals import from future causes makes the mode kwarg be of type unicode, and this breaks the numpy pad function when using python 2.7
But I agree that this really should not be there...

tacaswell: It would read better to put the `str` in the call to pad rather than in the signature.

lbluque: Right, reads way better!

PR #175, opened 2016-03-16
Small fixes

dmpelt:
Small fixes for #162.

I don't know why `map_async` was changed to `map` in #162, so I reverted it in e075779. If there was a good reason to do this, I can remove that commit. The reason for the `if p._pool` check is because `_pool` is undocumented, so maybe it will not work with future versions of `multiprocessing`. In these cases, we would still want to run a `map_async`, just one without the polling of the pool state.

### DIFF
--- a/tomopy/util/mproc.py
+++ b/tomopy/util/mproc.py
@@ -165,7 +165,7 @@ def distribute_jobs(arr,
 
     # if nchunk is zero, remove dimension from slice.
     map_args = []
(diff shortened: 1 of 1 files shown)


dgursoy:
Right. Could be my bad in merging.

dmpelt:
Did you mean to close this without merging? I won't mind if you just want to fix it in master directly.

dgursoy:
Ops.. Sorry. Haven't yet had my morning espresso..

PR #176, opened 2016-03-16
Fix ASTRA reconstruction

dmpelt:
This PR fixes ASTRA reconstruction after #162. 

I still need to check for performance regressions, so maybe we should wait on those before merging.

### DIFF
--- a/tomopy/recon/wrappers.py
+++ b/tomopy/recon/wrappers.py
@@ -79,7 +79,7 @@ needed_options = {
 }
 
 
-def astra(*args, on_host=False, **kwargs):
+def astra(*args, **kwargs):
     """
     Reconstruct object using the ASTRA toolbox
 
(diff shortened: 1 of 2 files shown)


dgursoy:
+1 on that. 

Can anyone fix the Travis build so that at least we have some tests before these merges? Yesterday I spent some time but for some reason installed Conda packages are not visible by the test scripts, and it yields import error. When I use pip for installing dependencies this seems to be avoided.

dmpelt:
I did some tests, and there do not seem to be any performance problems with this PR. There are also no significant performance improvements when using ASTRA, but that is probably expected since most time is spent computing things on the GPU anyways.

I am ok with merging this.

dgursoy:
Nice! Have you checked why this error shows up in Travis?

dmpelt:
Ah, I tested with Python 3 only, it seems that it does not work on Python 2 yet. Hooray for travis! I will try to fix this asap, probably tomorrow.

dmpelt:
Fixed PR for Python 2. Apparently, Python 2 does not support mixing `*args`, `**kwargs`, and actual keyword arguments that well, so I changed it to a global `mproc.ON_HOST` variable.

dgursoy:
Merge this?

dmpelt:
Fine with me! 

It could probably use a bit more testing by someone other than me, but I am pretty sure it will work.

dgursoy:
Ok, merging. Thanks!

PR #177, opened 2016-03-17
Fix travis builds by specifying conda path

dmpelt:
The old travis files did not specify the conda bin path in `PATH`, so the system `nosetests` command was used instead of the installed conda `nosetests` command. This PR fixes that.

There is still an error in Python 3.x versions, but that does not seem to be caused by the travis configuration.


dgursoy:
Thanks. This is good. Other builds fail because of py3 incompatibility and need fixes.

PR #181, opened 2016-03-22
Fix empty_shared_array for Python 3

dmpelt:
On Python 3, for some reason the `size` variable turned into a `numpy.int64` instead of a python `int`. 

`mp.RawArray` doesn't like `numpy.int64` as the input argument (it probably checks for `isinstance(size, int)` or something). Simply casting `size` to a python `int` fixes the problem and makes travis builds pass.


dgursoy:
Yay! All green on Travis!

PR #182, opened 2016-03-23
Moved exchange tasks out of tomopy

decarlof:
moved exchange/reader/writer out of tomopy to dxchange (http://dxchange.readthedocs.org)
doc update

to do: build dxchange and distribute it from the same channel of tomopy


decarlof:
travis fails because it cannot find the dxchange module ... how to add it?

dgursoy:
Then we should add the dxexchange to Conda recipe. Version at metadata and the VERSION file are not matching though.

decarlof:
I think I don't know how to add dxchange to conda recipe. I fixed the version and updated the conda build in my channel (https://anaconda.org/decarlof/dxchange).

dgursoy:
Do you think it makes sense to keep reader and writer here for generic load and saves?

tomopy:
1. Plus: Tomopy has independent ( = we can remove the dependencies from dxchange) basic r/w

2. Minus: other facility will be not aware of already written importer for their data set format.

I like 1.  and we can address 2. just linking the  dxchange documentation into the tomopy documentation. This can be under examples as it is now

F

Sent by T-mobile

dmpelt:
It might also be a good idea to keep the methods that already exist, but remove all code in them and make them raise an Exception with some information on how to install and use `dxchange`. This would, for example, be clearer for people that try out the example scripts of the paper.

tacaswell:
It might also be worth going through a deprecation cycle with all of those functions.  A warning + shimming to use the dxchange functions for a while and then fully removing.

dgursoy:
@decarlof could you merge with current master so that dxfile and dxchange are added to the travis.yml, and we can see if tests are passing or not?

data-exchange:
@dgursoy give me a few minutes ... the merge back with my master cannot complete automatically. It looks easy to solve (we both touched a the same file).

decarlof:
@dgursoy almost there ...

decarlof:
@dgursoy ok is good. Before merging though I want to take care of Daniel and Tom suggestions. Also we need to confirm if we still want to keep any basic r/w capability into tomopy at all

decarlof:
@dmpelt @tacaswell what about https://github.com/decarlof/tomopy/tree/deprecate

dmpelt:
@decarlof Looks good! I would maybe raise an exception after each warning as well, to make execution stop at that point instead of at a later point in the script, which might be confusing.

I noticed that the method names in `dxchange` are identical to the old ones in `tomopy`, so an alternative could also be something like the following in `tomopy.io.exchange`:

```
try:
    from dxchange import *
except ImportError:
    warnings.warn("TomoPy's I/O methods are deprecated and 
    will be removed in a future version. Please install the dxchange 
    module instead (see http://dxchange.readthedocs.org)",
         DeprecationWarning )
    
    [Old code implementing the methods + maybe a similar warning on each use]
```

This would have a couple of advantages:

1. Old scripts will keep working without having to change `tomopy` to `dxchange`.
2. Users that do not have `dxchange` installed will get a warning, but the scripts will run without error for a couple of TomoPy versions.
3. People that install `dxchange` will benefit from any improvements made in `dxchange`.
4. After a couple of version of TomoPy, everything except `from dxchange import *` can be removed without breaking any scripts. At this point it is also possible to deprecate the entire `tomopy.io.exchange` module for a couple of versions (showing warnings), and finally remove it entirely.

tacaswell:
@dmpelt suggestion is very user friendly, but I think just making `dxchange` a required dependency of tomopy would be enough. 

The shims in `exchange` need to call out to the same`dxchange` methods so that the functionality is not lost.

dgursoy:
How about using a decorator for function deprecations:
http://www.scipy-lectures.org/advanced/advanced_python/#deprecation-of-functions

prjemian:
Thanks for that link!

dmpelt:
@dgursoy That looks like a good solution!

decarlof:
- tomopy from now on will have a dependency from dxchange
- scripts using deprecated functions (i.e. tomopy.read_sls_tomcat()) will still work but get:
DeprecationWarning: Call to deprecated function write_tiff_stack. TomoPy's I/O methods are deprecated and will be removed in a future version (see http://dxchange.readthedocs.org)
- doc examples and tomopy.write_tiff in recon/rotation.py  is replaced with dxchange.write_tiff

meta.yaml
```diff
@@ -23,21 +23,12 @@ requirements:
     - scikit-image
     - pywavelets
     - pyfftw
```
dgursoy: Aren't we supposed to add dxchange here?

tomopy/io/exchange.py
```diff
@@ -83,7 +84,25 @@
            'read_petraIII_p05',
            'read_sls_tomcat']
 
-
+def deprecated(func):
```
dgursoy: I think this function should be placed somewhere inside util folder, since it's useful to any deprecated function in the future. But not sure where we should put it? ```util/misc.py```?

meta.yaml
decarlof: ops; added

tomopy/io/exchange.py
decarlof: not sure ether. I could  not find any 'best practices' recommendation, but it looks reasonable.  
Below is how it will look like

dgursoy: :+1:  Let's keep it as is for now. No need to complicate things. Thanks!

dgursoy: Ah. Already moved. Perfect!

PR #185, opened 2016-03-31
Do not build C library when running on RTD

### LINKED ISSUE #169: docs build needs to be corrected
After merging #162, readthedocs build keeps failing.

dmpelt:
This PR should fix #169 by skipping compilation of TomoPy's C code when installing on RTD systems. The C code is not needed for parsing the Python docstrings, so the documentation should still be complete. An `import tomopy` call complains a bit about missing external libraries, but imports fine, so RTD is able to build the documentation.


dgursoy:
Neat! I tested and works like a charm. Thanks!

PR #189, opened 2016-04-09
mock dxchange in RTD config

decarlof:
Docs failed to build API function links. This will fix it.

### DIFF
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -321,7 +321,7 @@ MOCK_MODULES = [
     'matplotlib', 'matplotlib.pylab', 'tifffile', 'EdfFile', 'netCDF4',
     'spefile', 'scipy.ndimage', 'pywt', 'scikit-image', 'skimage',
     'skimage.io', 'skimage.filter', 'skimage.morphology', 'skimage.feature',
-    'DM3lib', 'pyfftw']
+    'DM3lib', 'pyfftw', 'dxchange']
 
 for mod_name in MOCK_MODULES:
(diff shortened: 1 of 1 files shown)


doc/conf.py
```diff
@@ -321,7 +321,7 @@ def __getattr__(cls, name):
     'matplotlib', 'matplotlib.pylab', 'tifffile', 'EdfFile', 'netCDF4',
```
dgursoy: How about we also clean other file format packages from the list maybe?

decarlof: that has to stay till we keep the deprecated functions.

tomopy: Ah.. Can we simply put pass inside these functions and keep their name for some time?

decarlof: pass will force people to change their scripts; the current arrangement will send a warning but still work. I don't think we have  a big crowd using these importers so we can go with the pass, the only concern I have is the astra/tomopy paper that is just coming out with example using the older call; for this is probably worth it to keep the transition in 2 steps. Perhaps @dmpelt has some comments.

dmpelt: I would rather keep them for a couple of versions as well because of the paper. 

As an alternative for the docs, is it also possible to remove the `.. automodule::` things from the relevant `tomopy.io` module docs and insert a link to http://dxchange.readthedocs.org/ instead? I am not sure whether this will remove the need to specify the other file format packages in this list, but maybe a small change in the importing code could make that possible as well.

Issue #194, opened 2016-04-21
prep for 1.0.0


dgursoy:
We need to finish some work before we put the tag and put it in conda. 

To-do list:
- Update release notes: https://github.com/tomopy/tomopy/blob/master/doc/source/release/notes-1.0.0.rst: 
  - ASTRA addition as a new feature  
  - dxchange separation and deprecated features (also newly added file formats can be featured)
  - FFTW implementation as a new feature (need to close #178 at some point)
  - Gridrec modernization and backward incompatible changes
  - `minus_log`  #183 
- Add demo examples: https://github.com/tomopy/tomopy/tree/master/doc/source/demo
  - for the use of new filters for Gridrec.
  - for the use of ASTRA through TomoPy.
  - benchmarking script used in the paper 
- Update publication list: https://github.com/tomopy/tomopy/tree/master/doc/source/bibtex
- Few more bug fixes that I need to deal with:
  - downsample/upsample should work for number not in powers of two #188 
  - sino_360_to_180 bug

Please feel free to add/remove your names in the contributors list. This is a list of people who changed something one way or another.

PR #197, opened 2016-04-22
prep for 1.0.0

dgursoy:
#194


decarlof:
What about removing all examples from the docs (the same are available under DXchange) and replace them with Jupyter notebooks? Here is an example of how it will look like:

http://mytomopy.readthedocs.org/en/latest/ipynb/tomopy.html

dgursoy:
Will do. And since tomopy is file format independent we don't need to have multiple reconstruction examples of different facilities.

PR #196, opened 2016-04-22
Additions for debugging

dmpelt:
This PR adds two things to help debug problems:

* TomoPy will now reraise exceptions that are thrown by child processes, which will help debug problems like #195. Before, exceptions in child processes would simply result in a zero array being returned without any error message. Now it will show an error message, which might help a user to find out what went wrong (on Python 2 the message is quite short, but on Python 3 it will show a full traceback).
* A debug mode is added (turn on by calling `tomopy.set_debug()`), which will make everything run without any `multiprocessing`, but will still use the same chunk size and everything. This can help to find problems causing segfaults, since they will now cause the whole python instance to close instead of only stopping a single child process without a useful error message. Since it is implemented in `mproc`, it will work with any code that uses `mproc`. Obviously using debug mode will be much slower, since it will only use a single core.

Before first point (will return empty array):
```python
rec = tomopy.recon(d, ang,  algorithm=tomopy.astra, options={'method':'FBP', 'proj_type':'WRONG'})
```
After first point:
```python
rec = tomopy.recon(d, ang, ^J    algorithm=tomopy.astra, options={'method':'FBP', 'proj_type':'WRONG'})
---------------------------------------------------------------------------
RemoteTraceback                           Traceback (most recent call last)
RemoteTraceback: 
"""
Traceback (most recent call last):
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/multiprocessing/pool.py", line 119, in worker
    result = (True, func(*args, **kwds))
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/multiprocessing/pool.py", line 44, in mapstar
    return list(map(*args))
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/site-packages/tomopy/util/mproc.py", line 252, in _arg_parser
    result = func(*func_args, **kwargs)
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/site-packages/tomopy/recon/wrappers.py", line 122, in astra
    astra_run(*args, **kwargs)
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/site-packages/tomopy/recon/wrappers.py", line 170, in astra_run
    pi = astra_mod.create_projector(opts['proj_type'], proj_geom, vol_geom)
  File "/export/scratch2/pelt/condas/c3/lib/python3.5/site-packages/astra/creators.py", line 558, in create_projector
    return projector.create(cfg)
[...]

PR #203, opened 2016-05-12
Fixes multiprocessing jobs hanging when using remove_stripe_fw and retrieve_phase

lbluque:
I tested this a few times and had not hanging jobs. We might need to edit the `pyfftw.interfaces.cache.set_keepalive_time(5)` line or remove the caching altogether if hanging still occurs, but this seemed to work well. My idea is that the cache was being cleared before the multiprocessing jobs had been allocated and therefore hanging occured. Thanks to @HarinarayanKrishnan for catching this!

Also in this PR are some overdue unit tests for normalize_nf and find_center_pc


dgursoy:
@dmpelt I remember you also experienced this and suggested a fix like this before. Any comments on this?

tomopy/prep/phase.py
```diff
@@ -137,26 +138,20 @@ def retrieve_phase(
...
         proj = np.real(pyfftw.interfaces.numpy_fft.ifft2(
-            filtproj, planner_effort=_plan_effort(num_jobs))
+            filtproj, planner_effort=fftw_effort)
             ) / phase_filter.max()
         if pad:
             proj = proj[px:dy + px, py:dz + py]
         tomo[m] = proj
 
 
-def _plan_effort(num_jobs):
```
dgursoy: Hi Luis - why did you remove this function? Isn't it safe to have it once and call if needed, e.g. phase retrieval and stripe removal?

lbluque: I switched the function to a statement: `fftw_effort = 'FFTW_MEASURE' if num_jobs > 10 else 'FFTW_ESTIMATE'` (line 141 in phase.py and 132 in stripe.py).
This was to simplify the code (reduce function call and imports).
But you have a point that in terms of maintenance it is better to have it in one place.

dmpelt:
I don't remember having problems with hanging in TomoPy due to `pyfftw` caching, but I did increase the timeout in some of my personal codes.

I am a bit confused why the code is hanging because of this: I would expect each process to have a separate cache, and that `pyfftw` would be able to run correctly even if the cache is cleaned. Maybe we can try the new `tomopy.set_debug` functionality of #196 to investigate why it is hanging? It might not hang in that case though.

Also, do we actually need to enable the cache? The documentation [here](https://hgomersall.github.io/pyFFTW/pyfftw/interfaces/interfaces.html#module-pyfftw.interfaces.cache) suggests that caching only helps when doing a large number of (very) small transforms, which might not be relevant for typical TomoPy usecases. Is it easy to perform some computation time tests to see whether caching actually has a significant impact? If not, simply disabling the cache might solve the hanging problem.

Slightly related: maybe it is also useful to generate and save wisdom files after doing a `'FFTW_ESTIMATE`' call. I can imagine that users will often perform FFTs of identical sizes (e.g. the detector size of their experimental setup), in which case the wisdom file will help to speedup the first FFT significantly.

lbluque:
I agree with @dmpelt that we could disable the caching altogether. I did some quick comparisons caching vs non-chaching and speed differences were not very different. Plus disabling the cache also solves the hanging problem.

xianghuix:
Is this the issue related to #120?


dmpelt:
@xianghuix yes, I think they could be related. In general, if something crashes in one of the child processes, the issue of #120 will occur. This is a bit difficult to fix fully, but in recent Python versions (>3) and with recent commits to TomoPy things should behave a bit better (i.e. crash and report an error instead of hanging).

dgursoy:
What's the status of this PR? Should we still keep it open or is it already resolved?

tomopy/prep/phase.py
lbluque: @dgursoy should I switch this back to the original function call?

lbluque:
Well, I think we should get rid of the pyfftw caching like @dmpelt mentioned.
In terms of #120, perhaps this [http://scikit-learn.org/stable/faq.html#why-do-i-sometime-get-a-crash-freeze-with-n-jobs-1-under-osx-or-linux](url) on scikit-image might be related.

lbluque:
I removed pyfftw caching altogether.
@dgursoy Do you prefer to keep the `phase._plan_effort` function as it was before or keep the statement in this PR? This should be the last thing to address before merging.

dgursoy:
I would keep ```_plan_effort``` function. Then the PR is ready to be merged. Thanks.

lbluque:
Done.

dgursoy:
Thanks. Any idea why the Travis is failing?

lbluque:
Ah there was an error in the unit tests because I forgot to switch back the `_plan_effort` call in `_retrieve_phase`

dgursoy:
All right, thanks!

PR #213, opened 2016-07-12
Add ASTRA example to documentation

dmpelt:
This PR adds an example of how to use ASTRA through tomopy in the documentation. It is closely based on the existing tomopy gridrec example. The built example should show up on http://tomopy-dmpelt.readthedocs.io/en/astra_doc/ once it has correctly passed the build stage of RTD.


dgursoy:
👍

Issue #132, opened 2015-10-28
different results from legacy and current tomopy


rh66:
Just start to try tomopy in conda, and I get different results by using current tomopy in conda and legacy tomopy. I probably just don't know some fundamentals and therefore make the mistake. Hope someone can point out what I did wrong here.
I tested by assuming an object of round disc with intensity of 1 and radius of 200, with image width of 800. All the projections will be same since the rotation is centered at disc center. I generated projections, reconstructed it and plot the cross-section through disc center, as following:

```
data2=np.zeros((1200, 1, 800))
for i in range(400):
    data2[:, 0, i+200]=2*(200**2-(199.5-i)**2)**0.5
angle2=np.linspace(0,180.0,1200)*np.pi/180.0
rec2 = tomopy.recon(data2, angle2, center=399.5, algorithm='gridrec')
plt.plot(rec2[0,400,:])
plt.show()
```

The result is the left plot of attached image. If I use legacy tomopy, the result is shown on the right of attached image. The right one is my original object. And the code of using legacy tomopy is:

```
data2=np.zeros((1200, 1, 800))
for i in range(400):
    data2[:, 0, i+200]=2*(200**2-(199.5-i)**2)**0.5
dark = np.zeros((1, 800), dtype=float)
white = np.ones((1, 800), dtype=float)
theta = np.linspace(0, 180, 1200)
d = tomopy.xtomo_dataset(log='debug')
d.dataset(data2, white, dark, theta)
d.center=399.5
d.gridrec(emission=True)
```

Thank you for your attention and help.
Rong

![reconstructioncompare](https://cloud.githubusercontent.com/assets/6943918/10804812/e6103476-7d9f-11e5-99e7-e544c710f111.jpeg)

dgursoy:
It is likely because of different filtering, and interpolation values. I will check what the default values were before and after. Otherwise the algorithm is unchanged and should produce the same results.

dgursoy:
Tried with current version with the following code, and got some interesting results. 

- Top plot: Gridrec reconstruction of Shepp-Logan phantom. 
- Middle plot: Gridrec reconstruction values multiplied with two
- Bottom plot: Block-ART Iterative reconstruction plot

```Gridrec``` result seems to be multiplied by two. @michael-sutherland any comments or ideas why this may have happened?

```
import tomopy
import numpy as np

obj = tomopy.shepp2d(128)
ang = tomopy.angles(180)
prj = tomopy.project(obj, ang)

[...]

PR #216, opened 2016-07-14
Fix gridrec filter scaling

### LINKED ISSUE #209: Gridrec and FBP results are off from expected
We tested the tomopy recon algorithms with a phantom, and are not getting the expected results for pixel intensity values for gridrec and FBP (all other algorithms agree with what we expect). Gridrec pixel intensity values are off by a factor of approximately 1/2, and the FBP values are off by several orders of magnitude.

dmpelt:
This PR potentially fixes #209 and #132 by fixing the scaling of the gridrec filters. They seemed to be off by a factor of 2 in some cases. 

I basically compared with ASTRA's filter definitions, which includes the factor of 2 (https://github.com/astra-toolbox/astra-toolbox/blob/master/cuda/2d/fft.cu#L311). I tried to match the filter definitions of both TomoPy and ASTRA, but please check whether I made any mistakes.

I am definitely not sure about the Parzen and Butterworth filter definitions, as the Parzen definition in ASTRA seems quite different, and butterworth is not included in astra. Do you have specific mathematical definitions that you used when implementing them?


dmpelt:
I've activated the tests. I had to recreate the `.npy` data files, so at the moment it doesn't really test anything: the `npy` files were created with the same version. For future versions it will test whether something changed in the gridrec implementation.

dgursoy:
Ok, I will wait till npy  files are replaced then. For consistency, you can use proj.npy for the reproduction of the reconstructions.

dmpelt:
The new npy files should already be included in the PR (I used the `proj.npy` file to create them), so if you think I didn't make any mistakes I guess we can merge this. 

Of course, this PR does change reconstruction results for gridrec (which many people use), so should we include a notification or warning to users as well?

dgursoy:
I am merging this one but could you summarize changes in a list at: 
https://github.com/tomopy/tomopy/tree/master/doc/source/release/notes-1.1.0

It'll probably make our lives easier later on. Thanks!

PR #217, opened 2016-07-15
Performance improvements

dmpelt:
This PR uses `numexpr` to decrease computation time.

At the moment I applied it to `normalize` only, but I think the same approach could work on some other functions as well. I still have to check thoroughly whether it produces identical output, but the decrease in computation time seems worth it: for a dataset with 1025 projections with 128 slices and 2000 detector pixels, it reduces the time it takes to perform normalization from 1.61 s to 0.249 s on my machine. On another machine it reduces the normalization time from 4.15 s to 0.96 s for a 1025x256x2000 dataset.

I am not really sure what the cause is of the relatively slow performance of the current code on my machines: it could be that the `multiprocessing` module is doing some copying that we are not aware of. This could be related to the remarks at the end of #120.

Maybe the computation time of other functions (such as reconstruction) could be reduced by the same approach. I will investigate this further later, so we can wait with merging this until a later time. 

@dgursoy What are your thoughts on this?


dgursoy:
This is cool. I haven't used ```numexpr```  before, but this is a good excuse to give it a try. I'll also play with it and share my experiences. The only downside I can tell is that the code looks ugly, but that's fine I guess.

tomopy/prep/normalize.py
```diff
@@ -119,32 +120,23 @@ def normalize(arr, flat, dark, cutoff=None, ncore=None, out=None):
...
-# in-place normalization
-def _normalize(proj, flat, dark, cutoff):
-    denom = flat - dark
-    denom[denom < 1e-6] = 1e-6
-    proj -= dark
-    np.true_divide(proj, denom, proj)
+    flat = np.mean(flat, axis=0, dtype=np.float32)
+    dark = np.mean(dark, axis=0, dtype=np.float32)
+    
+    if ncore:
```
dgursoy: Do you need this ```if``` statement?

dmpelt: I wasn't sure what would happen if you call `set_num_threads` with `None` as argument. I just checked: it throws a `TypeError`. So we could keep the `if` or change it to a `try...catch` construction. Do you prefer one over the other?

dmpelt:
It looks a bit ugly indeed, but I think it is still readable, even if someone is not familiar with numexpr. 

We can remove the `local_dict=` things probably. By specifying a local_dict you get slightly better performance since it doesn't have to search the actual larger local dictionary, but in this case the benefits are probably minimal, and the code will look better if we remove it.

What do you think?

prjemian:
suggest this change:

    if ncore:

to

    if ncore is not None:

dmpelt:
I normally don't mind `if ncore` like checks for `None`, but an explicit check for `None` is indeed clearer. I could go either way.

I updated the PR to include `minus_log` as well. The previous `minus_log` implementation did not do any multithreading, so the improvements here are even greater: for a 1025x256x2000 dataset it reduces `minus_log` time from 5.232 s to 0.175 s.

dmpelt:
For tomopy reconstruction algorithms, we don't actually need the `multiprocessing` module, since the `ctypes` module drops the Python GIL, which makes threading with `threading` or `concurrent.futures` possible.

I've just added a commit to this PR that uses `concurrent.futures` instead of `multiprocessing` for reconstruction. On a dataset of size 1025x256x2000, this reduces the computation time for gridrec from around 21 seconds to about 9 seconds on my machine. I think the reason for this is mostly that it avoids copies of the data to `/tmp/` that `multiprocessing` seems to do.

This commit does break ASTRA integration I think: I will try to fix that in a later commit.

xianghuix:
Daniel,

This sounds very promising. I don't know the details but I heard the hanging issue with gridrec (#120) is related to multiprocessing. I am guessing your change will solve that issue as well? Is it correct?

Thanks,
x

dmpelt:
Yes, after this commit the gridrec code does not use multiprocessing anymore, so the hanging issue should be fixed. It might now crash instead in these cases, which will at least give us an easier way of diagnosing the underlying problem.

xianghuix:
That sounds great. Look forward to the new commit.

Best,
x

tomopy/misc/morph.py
```diff
@@ -121,15 +125,56 @@ def pad(arr, axis, npad=None, mode='constant', **kwargs):
...
+                            out=out[slc_l])
+                ne.evaluate("vec", local_dict={'vec': arr[slc_r_v]},
+                            out=out[slc_r])
+    else:
+        # Datatype not supported by numexpr, use numpy instead
+        out[slc_in] = arr
+        if mode == 'constant':
+            out[slc_l] = kwargs['constant_values']
+            out[slc_r] = kwargs['constant_values']
+        else:
```
lbluque: @dmpelt does this still support padding with the edge values of the array?

dmpelt: Yes, the `else` on line 144 and 155 handles the edge value case. It is not very readable, but in `_get_slices` the `slc_l_v` and `slc_r_v` variables make vectors on the left and right edge of the original array, and they get copied to the left and right padded area in the `else` cases.

lbluque: perhaps it will be more clear to have an `mode == 'edge'` and raise an error if kwarg `mode` is not constant, edge or whatever else is allowed.

dmpelt: Done!

dmpelt:
I think this can be merged, unless you want to test it a bit more first.

There are a few other routines that I might try to speed up a bit in the future, but I can create new PRs for them if needed.

dgursoy:
Thanks! Before we forget, could you write few items for the next release notes:
https://github.com/tomopy/tomopy/tree/master/doc/source/release

PR #223, opened 2016-07-26
Angle-dependent gridrec filters

dmpelt:
This PR enables the use of angle-dependent filters in TomoPy's gridrec implementation. It also adds the possibility to use a numpy array to specify such a filter (using `filter_name='custom2d'`). This will enable implementing recent filter-based algorithms easily with gridrec.

### DIFF
--- a/tomopy/recon/algorithm.py
+++ b/tomopy/recon/algorithm.py
@@ -140,6 +140,12 @@ def recon(
             Parzen filter.
         'butterworth'
             Butterworth filter.
+        'custom'
+            A numpy array of size `next_power_of_2(num_detector_columns)/2`
+            specifying a custom filter in Fourier domain.
+        'custom2d'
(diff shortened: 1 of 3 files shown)


tomopy/recon/algorithm.py
```diff
@@ -140,6 +140,12 @@ def recon(
             Parzen filter.
         'butterworth'
             Butterworth filter.
+        'custom'
```
dgursoy: It would be nice to explain if DC term is at the beginning of the array or at the middle? (e.g. if ```fftshift``` is needed or not for the filters.)

dgursoy:
I think this is ready to ship with a minor comment addition, right?.

dmpelt:
Yes - I will update it right away.

dmpelt:
Done. I think it can be merged now.

PR #225, opened 2016-08-18
find_center_vo bugfix + adding some test data

### LINKED ISSUE #224: find_center_vo() complaining about negative values
I'm attempting to use the find_center_vo() function to get the center of rotation for some data.  
When I run the function on my stack of projections without any optional values it gives the error "ValueError: negative dimensions are not allowed.".  

The same error is shown when I test the function using the sinogram.npy test data.  It makes it through the search_coarse() function, but gets hung up in search_fine() -> create_mask() with Ncol = -201.  

I'm not sure precisely what this algorithm is expecting for input.  My data doesn't have a full 180 degrees of projections so it makes sense that I'd get something odd from it, but I'm surprised that the test sinogram isn't working.  Could anyone shed some light on what could be going on?

skylarjhdownes:
Fixes #224


decarlof:
@skylarjhdownes please fix git identity on the local development machine. As it is now, any accepted pull request shows up as a commit contributed by "unknown"

skylarjhdownes:
I've set my identity locally, thanks for pointing it out.

PR #226, opened 2016-08-19
Fixes for sino_360_to_180

dmpelt:
This PR includes some fixes for `sino_360_to_180`, specifically:

* Fix output for `rotation='right'`
* Use both arrays for filling the overlapping region instead of only one array. This avoids using values at the edges of the detector, which may have the most artifacts in practice.
* Simplified the code a bit, using only a single `if` statement
* Renamed `sino_360_t0_180` to `sino_360_to_180` (`sino_360_t0_180` is kept for backward compatibility)
* Changed `rotation is` to `rotation ==`, using an equality comparison instead of an identity comparison (which often gives the correct results due to string caching, but can fail)
* Added a small test for the method

### DIFF
--- a/test/test_misc/test_morph.py
+++ b/test/test_misc/test_morph.py
@@ -68,6 +68,14 @@ def test_downsample():
 def test_upsample():
     loop_dim(upsample, read_file('obj.npy'))
 
+def test_sino_360_to_180():
+    test_im = np.random.random((32, 32, 128)).astype(np.float32)
+    ltest = sino_360_to_180(test_im, 32, 'left')
+    rtest = sino_360_to_180(test_im, 32, 'right')
+    assert_array_almost_equal(ltest[:, :, -112:], test_im[:16, :, 16:])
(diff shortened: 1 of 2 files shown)


dgursoy:
Still WIP?

dmpelt:
We have some experimental data that we want to test it on first (we had some errors with it, so I put it back to WIP). I will check whether it works now on Monday.

dgursoy:
sounds good!

dmpelt:
Tests on experimental data did not show any problems, so I think this can be merged!

Based on the tests, I added a note to the docs that the last projections is discarded for odd numbers of angles.

test/test_misc/test_morph.py
```diff
@@ -68,6 +68,14 @@ def test_downsample():
 def test_upsample():
     loop_dim(upsample, read_file('obj.npy'))
 
+def test_sino_360_to_180():
+    test_im = np.random.random((32, 32, 128)).astype(np.float32)
```
tacaswell: maybe add a test that exercises the odd number of frames behavior?

PR #229, opened 2016-09-14
Working to make the test suite more consistent.

skylarjhdownes:
I've made some sweeping changes to the structure of our unittests.  For consistency's sake I went with the class based unittest standard since that should work with any test runner.  Might be worth discussing whether we prefer that or the nose/py.test standard.


dmpelt:
Nice!

Nose is not actively developed any more, so I think it makes sense to switch to something else. The nose developers recommend "Nose2, py.test, or just plain unittest/unittest2" (see [here](http://nose.readthedocs.io/en/latest/)). Since you already put in the work for `unittest`, I think it is best to start using that. Are there any technical limitations to using `unittest` compared with other options?

skylarjhdownes:
The base library available in Python, `unittest` doesn't seem to have as feature rich a test runner as the other available libraries - test discovery specifically is much clunkier.  I suspect we'd be better off continuing to use `nose` than we would trying to revert to `unittest`, but I'm leaning towards using `py.test` since it seems to be the most actively maintained.  Both `nose` and `py.test` support unittest style tests.

My only hangup with `py.test` is that it doesn't want to run using the local source by default, instead testing against the installed library.  Only way I've found that it will test my local code is to run `python -m py.test` in the top level of the project.

dgursoy:
Is ```Pull Request Review``` a new thing in GitHub or was I ignoring it before?

skylarjhdownes:
Looks new to me: http://venturebeat.com/2016/09/14/github-launches-a-trello-competitor-pull-request-reviews-redesigned-profile-pages/

skylarjhdownes:
So the appropriate way to use py.test, or any Python test runner, appears to be to install your project in development mode so that the "installed" package is just a symlink to your development code.  Probably obvious to seasoned Python devs, but it was news to me.  If you `python setup.py develop` or `conda develop` to install, then `py.test` will run against your live code just fine.  

Is there anything preventing this from being merged at this point?

dgursoy:
@skylarjhdownes is this ready to go?

skylarjhdownes:
@dgursoy I think so.

PR #230, opened 2016-09-14
ASTRA wrapper fixes

dmpelt:
Fix for running ASTRA CPU algorithms that require extra options. This error was caused by my changes in #217 .

### DIFF
--- a/tomopy/recon/wrappers.py
+++ b/tomopy/recon/wrappers.py
@@ -207,6 +207,8 @@ def astra_rec_cpu(tomo, center, recon, theta, vol_geom, niter, proj_type, opts):
     import astra as astra_mod
     nslices, nang, ndet = tomo.shape
     cfg = astra_mod.astra_dict(opts['method'])
+    if 'extra_options' in opts:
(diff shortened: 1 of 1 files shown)


dmpelt:
PS. No idea why travis fails with 3.4. It seems to be a conda/travis problem, since the current master fails for me at this time too. The #229 build did not fail, so something in conda/travis changed today apparently.

dgursoy:
Not sure we want that, but would adding ```--copy``` flag to conda may solve it? It will then install all packages using copies instead of hard- or soft-linking.

But if could also be a bug in Travis since I didn't understand why it downgrades from Python 3.5.2 to Python 3.4.5-0 (and at the beginning it said it has Python 3.4.2).

dgursoy:
@dmpelt what is the status of this PR?

dmpelt:
I think it is actually working, but Travis was having some problems that day. Is there a way to force a Travis rebuild without pushing some random commit to it?

dgursoy:
Restarted builds. Only 3.4 was failing.

dgursoy:
Seems good to go!

PR #232, opened 2016-10-18
Add instructions to use TomoPy with UFO

matze:
I updated our TomoPy [wrapper](https://github.com/ufo-kit/ufo-core/blob/08ad1812e7522026ea1e1c4596235339d510eb61/python/ufo/tomopy.py) (still ironing out some issues with DFI) and re-ran your IPython notebooks with some adaptions. As it seems to work quite well, I'd appreciate if you could include this change to the documentation.

P.S.: To avoid inflating the repo size, it would be a good idea to use the same image file for those results that do not change.


doc/source/ipynb/ufo.rst
```diff
@@ -0,0 +1,126 @@
+
+Reconstruction with UFO
+-----------------------
```
dgursoy: This is neat! Can you also add one or two sentences at the beginning for explaining what UFO is, who develops it, and the system and processor requirements?

doc/source/ipynb/ufo.rst
```diff
@@ -0,0 +1,126 @@
...
+
+    plt.imshow(recon[0, :,:], cmap='Greys_r')
+    plt.show()
+
+
+
+.. image:: ufo_files/ufo_26_0.png
+
+
+For faster reconstruction, we can use UFO's Direct Fourier Inversion
```
dgursoy: Can you add reference to the algorithm and/or to its implementation? 
You can update the bibtex file by adding UFO related papers: 
https://github.com/tomopy/tomopy/tree/master/doc/source/bibtex

doc/source/bibtex/ref.bib
```diff
@@ -178,6 +178,30 @@ @article{Toby:15
...
+volume = {9967},
+pages = {996715-996715-9},
+year = {2016},
+doi = {10.1117/12.2237611},
+}
+
+@inproceedings{vogelgesang2012ufo,
+author = {Vogelgesang, M. and Chilingaryan, S. and Rolo, T.  dos Santos and Kopmann, A.},
+booktitle = {Proceedings of The 14th IEEE Conference on High Performance Computing and Communication \& The 9th IEEE International Conference on Embedded Software and Systems (HPCC-ICESS)},
+title = {UFO: A Scalable GPU-based Image Processing Framework for On-line Monitoring},
```
dgursoy: You can rather have this in direct citations rather than references:
https://github.com/tomopy/tomopy/blob/master/doc/source/bibtex/cite.bib
and add:
...and also cite [A2] if you use ASTRA, and [A3] if you use UFO algorithms).
Please let me know before merge.

matze:
Thanks for the additional information. I (force) pushed a squashed commit, so you can merge if you like (as an external user I cannot :-) )

dgursoy:
Looks good! Thanks!

dgursoy:
RTD is complaining. Can you please check if the bibtex style is correct:
http://readthedocs.org/projects/tomopy/builds/4527651/

matze:
Hmm, this is strange. It complains that it cannot find `booktitle` but it is defintely [in](https://github.com/tomopy/tomopy/blob/master/doc/source/bibtex/ref.bib#L184).

dgursoy:
Ah.. Nevermind. I just added it.

PR #234, opened 2016-10-25
add tomocuda interface

kyuepublic:
The new change are providing cuda support for median_filter and outlier_removal.


dgursoy:
Tests seems to be failing.

kyuepublic:
I checked the failing information. So it needs to install my tomocuda package on the travis server.... My package can be download here, https://github.com/kyuepublic/tomocuda. 

Also, it will need cuda library to be installed on this server.

dgursoy:
I believe installing on this on Travis CI is against the principles of continuous integration, and even not sure if we can do that. Before we merge you need to have ```travis.yml``` file edited so that the dependency can be updated to Travis through Conda. You can send me your conda recipe for tomocuda and its dependencies and I can upload your package to the tomopy channel so that Conda package manager finds everything it needs for installation.

tomopy/misc/corr.py
```diff
@@ -277,7 +323,7 @@ def remove_neg(arr, val=0., ncore=None):
 
 def remove_outlier(arr, dif, size=3, axis=0, ncore=None, out=None):
     """
-    Remove high intensity bright spots from a 3D array along specified
+    Remove high intensity bright spots from a 3D array along 0 axis
```
tacaswell: This seems to be at odds with the rest of the docstring and (I think) the code.

tomopy/misc/corr.py
```diff
@@ -188,6 +191,49 @@ def median_filter(arr, size=3, axis=0, ncore=None):
...
+    size : int, optional
+        The size of the filter.
+
+    Returns
+    -------
+    ndarray
+        Median filtered 3D array.
+    """
+
+    winAllow = [2,3,4,5,6,7,8,9.10,11,12,13,14,15]
```
tacaswell: this would be marginally be better as a set.

tomopy/misc/corr.py
```diff
@@ -323,6 +369,64 @@ def remove_outlier(arr, dif, size=3, axis=0, ncore=None, out=None):
...
+            im_noisecu=np.lib.pad(im_noisecu, ((loffset, roffset),(loffset, roffset)), 'symmetric')
+            im_noisecu = im_noisecu.flatten()
+
+            filter.setCuImage(im_noisecu)
+            filter.run2DRemoveOutliner(size, dif)
+            results = filter.retreive()
+            results=results.reshape(imsizey,imsizex)
+            out[step]=results
+    else:
+        print("using cpu remove outlier")
```
tacaswell: rogue print

tomopy/misc/corr.py
```diff
@@ -61,6 +61,7 @@
 import logging
 import numexpr as ne
 import concurrent.futures as cf
+import tomocuda
```
tacaswell: Better to move this inside of the function call and gracefully fall-back to cpu if tomocuda can not be imported.

tomopy/misc/corr.py
dmpelt: A `size>1 and size<16` would work as well I think?

tomopy/misc/corr.py
tacaswell: probably also do a check on the GPU available and fall back to cpu if a usable one is not found?

tomopy/misc/corr.py
kyuepublic: Hi tdmpelt:

the window size=1 and size>15 is not supported. Actually, GPU performs worse and worse when you increase the window size. 

Thanks

kyueaes:
Hi Goga:

I do have a conda recipe for tomocuda. But that is for cuda 7.5. The problem I found right now is that, if the cuda version is not the same, there will be issue with the package. So the best way to do is to compile from the source.

Regards,
Ke Yue

tomopy/misc/corr.py
dmpelt: Yes, I understand. I mean that the `if` statement can be changed to `if size>1 and size<16:`, which is more readable (to me at least), and faster as well.

tomopy/misc/corr.py
```diff
@@ -188,6 +195,57 @@ def median_filter(arr, size=3, axis=0, ncore=None):
...
+                im_noisecu=np.lib.pad(im_noisecu, ((loffset, roffset),(loffset, roffset)), 'symmetric')
+                im_noisecu = im_noisecu.flatten()
+
+                filter.setCuImage(im_noisecu)
+                filter.run2DFilter(size)
+                results = filter.retreive()
+                results=results.reshape(imsizey,imsizex)
+                out[step]=results
+        else:
+            print("The window size is not support, using cpu median filter")
```
tacaswell: These should be warning, not print statements.

tomopy/misc/corr.py
```diff
@@ -62,6 +62,11 @@
 import numexpr as ne
 import concurrent.futures as cf
 
+try:
+    import tomocuda
```
tacaswell: Do not really need this block as tomocuda is always imported locally below?

dgursoy:
Thanks @kyuepublic. It looks great and almost ready for the merge. Could you please make the PEP8 corrections, and add if possible the axis argument before merge?

Issue #221, opened 2016-07-20
ncore/nchunk having no effect


cmkewish:
I'm seeing no difference in run times of gridrec for various versions of the parameters ncore and nchunk, using an edited version of the TomoPy example script on readthedocs.

Version: TomoPy 1.0.0
System: 4x 16-core CPU, 4TB RAM
OMP_NUM_THREADS=1 in system environment
Dataset: 1500 projections x 2048x2048 frame size x 17 MB TIFF
Read data function call:

```
for i in range(0,1499):
    file = '%sextpro_%04d.tiff' % (fname, i)
    proj[i,:,:] = dxchange.read_tiff(file)
```

Reconstruction function call: 

```
recon = tomopy.recon(proj, theta, center=1024, algorithm='gridrec', ncore=64, nchunk=32)
```

Output dataset: 2048x 33 MB TIFF

Writing in /home
ncore=64 and nchunk=32: `
    real    12m22.079s
    user    14m47.912s
    sys     5m59.012s`
run 2`
    real    12m21.421s
    user    15m2.416s
    sys     5m52.600s`
ncore=16 and nchunk=128`
    real    12m40.706s
    user    11m54.890s
    sys     5m26.917s`

Using a different write directory on a mount point:
ncore=64, nchunk=None`
    real    11m39.197s
    user    15m20.341s
    sys     5m37.033s`
2nd run:`
    real    10m25.411s
    user    15m24.228s
    sys     5m24.221s`
ncore=4, nchunk=None`
    real    12m20.970s
    user    10m5.310s
    sys     5m2.371s`
ncore=4, nchunk=1`
    real    14m21.153s
    user    14m46.437s
    sys     5m8.720s`

Is there something I'm supposed to enable for these parameters to work?

dgursoy:
A bit of a side topic, but could you also check times by setting ```sinogram_order=True``` in your reconstruction? For this, you should first swap the order of the first and second axes of the sinogram before reconstruction:
```
for i in range(0,1499):
    file = '%sextpro_%04d.tiff' % (fname, i)
    proj[:,i,:] = dxchange.read_tiff(file)
```
```
recon = tomopy.recon(proj, theta, center=1024, algorithm='gridrec', sinogram_order=True)
```

PR #241, opened 2016-12-02
Fixed up chunking and multithreading.

michael-sutherland:
Fixed a bug where recon wasn't using the nchunk parameter (see #221 ), it now does chunking in a better way.  Also, cleaned up the extern module to always call the underlying function instead of sometimes returning a tuple containing the function and its parameters.  Further, fixed up the Astra wrapper to match my changes.  @dmpelt The astra wrapper doesn't appear to be supporting multiple GPUs, but I think this was already an issue.  Specifically, if I try to run on two GPUs and step through the code, the first one starts up and then is stopped when I submit the thread for the second one.

### DIFF
--- a/tomopy/misc/corr.py
+++ b/tomopy/misc/corr.py
@@ -146,13 +146,12 @@ def gaussian_filter(arr, sigma=3, order=0, axis=0, ncore=None):
     if ncore is None:
         ncore = mproc.mp.cpu_count()
 
-    e = cf.ThreadPoolExecutor(ncore)
-    slc = [slice(None)]*len(arr.shape)
-    for i in range(arr.shape[axis]):
-        slc[axis] = i
-        e.submit(filters.gaussian_filter, arr[slc], sigma, order=order,
-                 output=out[slc])
-    e.shutdown()
+    with cf.ThreadPoolExecutor(ncore) as e:
+        slc = [slice(None)]*arr.ndim
+        for i in range(arr.shape[axis]):
+            slc[axis] = i
+            e.submit(filters.gaussian_filter, arr[slc], sigma, order=order,
+                     output=out[slc])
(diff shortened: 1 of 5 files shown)


kyueaes:
What is the output command of this "nvidia-smi -s", Michael?

Regards,
Ke Yue

michael-sutherland:
I am running two Titan X (Pascal) GPUs.  Did you mean "nvidia-smi -a"?

kyueaes:
Hi Michael:

No, how about "nvidia-smi -q", what is the ouput? 

I guess it might because your gpu is not set to exclusive mode.

Regards,
Ke Yue

________________________________________

michael-sutherland:
I set to exclusive mode, but am seeing the same issue.  Here's the output from nvidia-smi -q.

[gpu_query.txt](https://github.com/tomopy/tomopy/files/628294/gpu_query.txt)

kyueaes:
umm...it might be some other issue then. If possible could you try set it to  DEFAULT and EXCLUSIVE_THREAD MODE to check if it is working?

Thanks,

michael-sutherland:
I had it on DEFAULT originally.  It won't let me set EXCLUSIVE_THREAD, it has apparently been depricated.

michael-sutherland:
If someone can verify that it works for them with 2 GPUs, I'll update my commit comment.  For now, moving the discussion to #242.

michael-sutherland:
Just fixed a bug in the poll request for when you specify nchunk (it was creating too many slices).

dgursoy:
@michael-sutherland Should this thread be kept open or ready to go?

michael-sutherland:
It should be ready to go.

dgursoy:
All right then. Thanks!

PR #244, opened 2016-12-08
Remove ring improvements

dmpelt:
The `tomopy.remove_ring` code does some filtering in polar space to smooth the estimated rings, after which they get subtracted from the reconstructed image. However, the current code smooths over the entire 360 degree range. In reality, however, most rings are actually half-rings, since acquisition is from 0-180 degrees. Therefore, at the edges of the half-rings, the rings would get 'smoothed' into the filtered image, which can cause significant artifacts (an example image is shown below).

In this PR, a new option is added to `tomopy.remove_ring`, called `int_mode`, which can be set to `'WRAP'` for the old behaviour, or to `'REFLECT'` for the new behaviour. In the new behaviour, the polar image is smoothed from 0-180 and 180-360 separately, using reflective boundary conditions at 0, 180, and 360 degrees. This reduces the artifacts, although there is still a 'jump' present (caused by the 'jump' in half-rings meeting at that point). See below for comparison images.

Note that it is probably best to let the user decide which half-rings to filter (for example for problems with limited angular range), but that will require larger changes to the code. Also, I removed some unused code from `remove_ring.c` while I was working on it.

Example (note, I just chose some random values for the parameters - it could probably be much better filtered, but I just want to show the differences):
Before:
ring_before
After:
ring_after

### DIFF
--- a/src/remove_ring.c
+++ b/src/remove_ring.c
@@ -45,10 +45,13 @@
 
 #include "remove_ring.h"
 
+#define INT_MODE_WRAP 0
+#define INT_MODE_REFLECT 1
+
 void remove_ring(
 	float* data, float center_x, float center_y, int dx, int dy,
 	int dz, float thresh_max, float thresh_min, float threshold,
-	int angular_min, int ring_width, int istart, int iend)
+	int angular_min, int ring_width, int int_mode, int istart, int iend)
 {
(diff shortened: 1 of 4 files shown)


dgursoy:
This is nice!

Issue #242, opened 2016-12-02
Issue running Astra wrapper with 2 GPUs


michael-sutherland:
Moving discuss from #241 to here.  I can't run with more than one GPU when using the Astra wrapper.  Stepping through the code, it looks like the first GPU starts up just fine, but the second GPU kicks off the first one.  I tried changing my compute mode to EXCLUSIVE_PROCESS but it didn't seem to help.

dmpelt:
I have been running ASTRA with multiple GPUs successfully in the past, so at least it was working before. I haven't tried it recently though, so maybe some of the recent changes broke it, although I am not seeing something suspicious in the code at the moment. I will try it soon on one of our machines here.

Have you tried running with `ncore=1` in the `tomopy.recon` call? Since all parallelization is done by the GPUs anyways, using `ncore=1` is usually better when running ASTRA CUDA algorithms. When `ncore>1`, TomoPy will launch multiple threads that each launch multiple threads themselves (one for each GPU) - maybe that causes issues...

michael-sutherland:
Yes, I was using ncore=1 (also, I just changed it so that we don't spawn a thread when ncore=1).

dmpelt:
That is strange then... I will test with the latest master branches of both TomoPy and ASTRA and report back.

dmpelt:
I just checked with the latest master branches of both TomoPy and ASTRA:

If I don't supply `ncore=1`, the code is only run on the first index of `gpu_list`. When I use `ncore=1` however, it does run on all GPUs in `gpu_list`. So the following code runs on four GPUs on my test machine:

```python
tomopy.recon(d,a,algorithm=tomopy.astra, options={'method':'FBP_CUDA','proj_type':'cuda','gpu_list':[0,1,2,3]},ncore=1)
```

I'm not sure why it doesn't work when not using `ncore=1`: it could be that the threads-within-threads are causing problems. I will try to find out what is going on. 

Also, I don't know why it is not running on both GPUs on your machine even when using `ncore=1`. What does `nvidia-smi` output before and after starting the reconstruction call?

michael-sutherland:
Before processing starts:
```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 370.28                 Driver Version: 370.28                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
[...]

PR #246, opened 2016-12-15
Fix for running on mulitple GPUs.

michael-sutherland:
There was a shared dict that was being modified with the GPU Index in different threads.  Fix is to deepcopy the dict before passing to different threads.

### DIFF
--- a/tomopy/recon/wrappers.py
+++ b/tomopy/recon/wrappers.py
@@ -57,6 +57,7 @@ import logging
 from tomopy.util import mproc
 
 import numpy as np
+import copy
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


michael-sutherland:
This fixes #242.

dmpelt:
Small comment: could we also do the deep copy at the `cfg['option'] = opts['extra_options']` line (i.e. `cfg['option'] = copy.deepcopy(opts['extra_options'])`)? In that case, we are only copying the `extra_options` dict, which is potentially faster than copying the entire `opts` dict (also, does `copy.deepcopy` create a deep copy of a dict within a dict?).

michael-sutherland:
copy.copy() is a shallow copy.  copy.deepcopy() will copy all of the elements recursively.  I debated putting the copy there as well, but decided to put it where the multithreading occurred (since astra_rec_cuda gets called in the multi-threaded and single-threaded case).  I'm not sure which is actually better from a readability and maintainability perspective.

michael-sutherland:
I thought about it last night and am going to change the copy to the place you specified.  We shouldn't be changing the parameters we pass in unless it is explicit and could also mess something up for the single threaded case in the future.  I'll switch where the change is happening and update the pull request.

dmpelt:
I was just about to comment that either way was fine with me :)

The PR looks good to be merged to me!

PR #248, opened 2016-12-21
Add 1D outlier removal and make current outlier removal support N dimensions

dmpelt:
This PR adds a new function, `remove_outlier1d`, that is similar to the existing `remove_outlier`, but performs median filtering only along a single axis. In addition, it makes a small change to the existing `remove_outlier` code such that is supports N-dimensional arrays (instead of only 3-D), and updates its documentation to clarify what is does.

We found the 1D outlier removal to be useful when only processing a small number of slices (for example just one). In these cases, the existing 2D outlier removal would use only a small filter size (since part of the filter is 'cropped' by the limited number of slices). Since there is usually some cross-talk between neighboring pixels, the small median filter would not remove many of the existing outliers. Using a larger filter size will work in these cases, but then the required filter size depends on the number of slices the user is processing.

Using the added `remove_outlier1d` function, it is possible to filter along the angle dimension in 1D (which is the default case). Since there is no cross-talk between projections, a small filter size correctly filters out most outliers, even when only processing a small number of slices. In fact, each slice is filtered independently from the others in this case, so the output of the procedure is identical when processing few slices compared to processing many slices.

While working on this PR, I noticed that `remove_outlier` only supported 3-D arrays: I have changed the code so that it supports generic N-D arrays. Furthermore, the documentation of `remove_outlier` was a bit confusing to me: it suggested that the filtering was performed along the `axis` axis, but the filtering is actually performed along all axes other than `axis`. I have tried to clarify this in the documentation.

### DIFF
--- a/tomopy/misc/corr.py
+++ b/tomopy/misc/corr.py
@@ -394,7 +394,7 @@ def remove_outlier(arr, dif, size=3, axis=0, ncore=None, out=None):
 def remove_outlier1d(arr, dif, size=3, axis=0, ncore=None, out=None):
     """
     Remove high intensity bright spots from an array, using a one-dimensional
-    median filter along specified axis.
+    median filter along the specified axis.
 
     Parameters
     ----------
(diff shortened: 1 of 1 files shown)


tomopy/misc/corr.py
```diff
@@ -375,21 +377,74 @@ def remove_outlier(arr, dif, size=3, axis=0, ncore=None, out=None):
...
                      output=tmp[slc])
 
     with mproc.set_numexpr_threads(ncore):
         out = ne.evaluate('where(arr-tmp>=dif,tmp,arr)', out=out)
 
     return out
 
+def remove_outlier1d(arr, dif, size=3, axis=0, ncore=None, out=None):
+    """
+    Remove high intensity bright spots from an array, using a one-dimensional
```
dgursoy: Does it make sense to add here that the 1-D median filter is always computed along horizontal axes of the 2-D image defined by axis argument?

dmpelt: I think the method works differently if I remember correctly: it computes the 1-D median filter along the axis defined by the axis argument, and slices along a different axis (the largest of the remaining axes). The other outlier methods indeed do the slicing in the direction of the axis argument, and perform the filtering along the other axes - this was confusing to me in the old method documentation (I only understood it by reading the code), so the PR also includes a clarification of this in the documentation.

Issue #249, opened 2017-01-06
RTD build fails


dgursoy:
Recently, RTD Sphinx build fails:

```
Running Sphinx v1.3.5
making output directory...

Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/tomopy/envs/latest/local/lib/python2.7/site-packages/sphinx/cmdline.py", line 243, in main
    opts.warningiserror, opts.tags, opts.verbosity, opts.jobs)
  File "/home/docs/checkouts/readthedocs.org/user_builds/tomopy/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 144, in __init__
    self.setup_extension(extension)
  File "/home/docs/checkouts/readthedocs.org/user_builds/tomopy/envs/latest/local/lib/python2.7/site-packages/sphinx/application.py", line 435, in setup_extension
    err)
ExtensionError: Could not import extension nbsphinx (exception: No module named jupyter_client.manager)

Extension error:
Could not import extension nbsphinx (exception: No module named jupyter_client.manager)
```

I haven't throughly checked but nbsphinx is already listed in the requirements-doc.txt. I suspect that there may be a compatibility issue with IPython?

decarlof:
it seems jupyter_client.manager is now part of jupyter.

PR #250, opened 2017-01-09
working on RTD build fails #249

decarlof:
the RTD build fails #249 issue is due to the fact that jupyter_client is now part of jupyter and not ipython => I replaced this in the requirements-doc.txt file but bumped into another issue. The new jupyter_client is using the | operator that I had to mock but this is not enough since the doc building now gives:
```
sys.setdlopenflags(ctypes.RTLD_GLOBAL | dlflags)
TypeError: an integer is required
```

does anyone know how to mock a function?


decarlof:
@dgursoy RTD now builds fine, you can merge this pull request. Please note I did sync with your (dgursoy/tomopy) fork before testing the doc built correctly.

Issue #236, opened 2016-11-01
Missing dependencies


kohr-h:
When `pip install`'ing straight from a repo clone, `import tomopy` fails due to a couple of missing dependencies. Perhaps you want to add them to `setup.py`?

List of missing dependencies:
- `pyfftw` (raised in `__init__.py`)
- `six` (raised in `util/dtype.py`)
- `h5py` (raised in `ip/writer.py`)
- `numexpr` (raised in `util/mproc.py`)
- `tifffile` (raised in `misc/phantom.py`)
- `dxchange` (raised in `recon/rotation.py`)
- `pywt` (raised in `prep/stripe.py`)
- `skimage` (raised in `recon/rotation.py`)

Edit: alternatively, if you want it to be optional, the potentially failing import statements should be fenced, and the case when they're not available should be handled.

Edit 2: Instead of opening new issues, I just add the other missing dependencies here.

carterbox:
If you use `pip` instead of `conda` to install, the [requirements](https://github.com/tomopy/tomopy/blob/master/requirements.txt) list has most of these already, but if you are installing from a local clone, then I think you have to use `-r requirements.txt` to request that the dependencies are installed.

Considering the above, it seems that only `h5py` and `tifffile` are missing, but these are optional. I'm not sure if they are actually needed anyway because I believe that all the file io operations are being handled through `dxchange` now/soon.

kohr-h:
Sure, it's up to you how you want to handle this. Requiring the `-r requirements.txt` is one way, but it's a bit annoying from a user's perspective. You can also read the file in `setup.py` and use the lines as `install_requires` in the call to `setup`.

carterbox:
It's not up to me; I'm not a maintainer for this repository.

I agree it's a little annoying, but the recommended/supported package manager for `tomopy` is `conda`. With `conda`, the dependencies are installed automatically.

kohr-h:
> It's not up to me; I'm not a maintainer for this repository.

Well then it's up to them if they want to close this issue or do something about it.



> I agree it's a little annoying, but the recommended/supported package manager for tomopy is conda. With conda, the dependencies are installed automatically.

That's true for releases, but for installation from source, the [instructions](https://tomopy.readthedocs.io/en/latest/install.html#installing-from-source-via-command-line) give an "old-fashioned" command (with `sudo`) that roughly does the same as `pip install --editable /path/to/repo`. So with the official instructions, you end up with an installation that is not usable, which is kind of bad.

[...]

PR #253, opened 2017-02-17
DOC: Clarification of install from source

dgursoy:
#236


decarlof:
@dgursoy looks good but docs still fails to build (see #250)

dgursoy:
Ah.. yes - thanks. @decarlof: Is #250 ready to go?

decarlof:
@dgursoy not yet I stock on the function mocking. Perhaps @prjemian can help?

PR #263, opened 2017-05-25
Bug fixes for windows

EricPythonMaster:
This PR primarily introduces a number of fixes for compiling and running on Windows.
Also, when the number of slices is less than the number of cores, do not create extra work for the unused cores.


EricPythonMaster:
I targeted the wrong branch.

PR #264, opened 2017-05-25
Bug fixes for windows

EricPythonMaster:
This PR primarily introduces a number of fixes for compiling and running on Windows.
Also, when the number of slices is less than the number of cores, do not create extra work for the unused cores.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,9 @@ else:
 extra_comp_args = ['-std=c99']
 extra_link_args = ['-lm']
 if os.name == 'nt':
+    import sys
+    if sys.version_info.major == 3:
+        extra_comp_args += ['-DPY3K']
     extra_comp_args += ['-DWIN32']
     extra_link_args += ['-lfftw3f-3']
 else:
--- a/src/gridrec.c
+++ b/src/gridrec.c
@@ -53,6 +53,8 @@
 #include <fftw3.h>
 #include <math.h>
 #include <string.h>
+#include <pthread.h>
+pthread_mutex_t lock;
 
 #ifndef M_PI
     #define M_PI 3.14159265359
@@ -75,7 +77,7 @@ gridrec(
     const int ltbl = 512;         
     int pdim;
     float _Complex *sino, *filphase, *filphase_iter, **H;
-
+    pthread_mutex_lock(&lock);
     const float coefs[11] = {
          0.5767616E+02, -0.8931343E+02,  0.4167596E+02,
         -0.1053599E+02,  0.1662374E+01, -0.1780527E-00,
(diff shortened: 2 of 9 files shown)


EricPythonMaster:
For those who follow this, the error that causes the tests to fail are not reproducible on windows, however, the bug with multi-threading still occurs through FFTW.

Somewhere along the way, fftw receives a nullptr and it causes it to crash.
I also tried to use the "fftw_make_planner_thread_safe" and it didn't resolve anything. This was tested on fftw 3.3.5 (it is known that in 3.3.5 the thread safe doesn't work) and there are no ffftw 3.3.6 binaries yet for windows.  I will have to compile it at some point, but this could be it looking at the gridrec code. If someone wants to build them and send them to me, that would be greatly appreciated.

It would be surprising if it was just that as it works fine on linux.
If there are bugs on linux, but  they aren't apparent, or no one has noticed them yet.
I will definitely keep working on this when I have some time.

EricPythonMaster:
For anyone wondering:
FFTW3.3.6 on windows 
https://anaconda.org/ccpi/fftw

EricPythonMaster:
I used a mutex in grid rec. That works now.
Multiprocessing also works now.

For some reason, project is not working.
Doing a projection creates a 0 valued matrix.

EricPythonMaster:
@dgursoy @dmpelt 
Alright, all of the test have passed now with minimal impacts on the rest of the code.
The biggest change is with fftw in gridrec. I've now wrapped it with a mutex. It was a problem for winidows, but apparently no obvious problems on linux. However, to prevent possible issues, I left the mutex for linux as well.

The multiprocessing on windows works as well. Here's a screenshot of the test passing on my windows machine.
windows test passed

The main issue with the multiprocessing was that it was not used properly. The numpy arrays were getting pickled between processes on windows. With this PR,I 've also noticed that the multiprocessing tests run faster, at least on windows.

dgursoy:
@EricPythonMaster I wonder if everything is fine in Windows after the current replacement of FFTW with mkl_fft. How easy one can write a AppVeyor CI script for Win?

thejinx0r:
@dgursoy 
This is my regular account. 

I have no experience with AppVeyor.
I'll have to look in to it more carefully, but a glance, the syntax is quite similar to Travis.

I think it would be quite easy to write something up for it.

dgursoy:
@thejinx0r I have no experience with Windows, and no way to test. I hope somebody can jump in and give it a try.

thejinx0r:
@dgursoy 
I'll give it a go.
Definitely something I wanted to add to the CI. I just never knew AppVeyor supported Windows.

PR #268, opened 2017-06-22
WIP: replace multiprocessing by threading

dmpelt:
This is still WIP, please do not merge yet!

I was working on replacing some remaining methods that use multiprocessing by threading, since it improved computation speed and memory requirements in other methods. For some methods (e.g. Fourier-stripe ring removal), it turns out that quite a lot of things have to be changed.

This PR still needs some work, for example testing and documenting. For now, I just took the changes from last year and merged them with the latest master branch. 

I've already opened the PR to document the progress and to make it possible for other to already comment on things


tomopy/prep/normalize.py
```diff
@@ -202,15 +206,22 @@ def normalize_bg(tomo, air=1, ncore=None, nchunk=None):
...
-        args=(air,),
-        axis=0,
-        ncore=ncore,
-        nchunk=nchunk)
-    return arr
+    chnks = np.round(np.linspace(0, axis_size, ncore+1)).astype(np.int)
+    mulargs = []
+    for i in range(ncore):
+        mulargs.append(extern.c_normalize_bg(out[chnks[i]:chnks[i+1]], air))
+    e = cf.ThreadPoolExecutor(ncore)
```
tacaswell: If you are going to use cf, it is probably better to allow the schedule to be passed in (and default to a new thread pool if none is provided).  This will allow something like `dask` to just 'drop in' or to re-use a thread pool across function calls.

tacaswell:
A decorator like

```python
def ensure_scheduler(func):
    @functools.wraps(func)
    def inner(*args, **kwargs):
         if 'sched' not in kwargs:
             kwargs['sched'] = cf.ThreadPool(ncores)
        return func(*args, **kwargs)

@ensure_scheduler
def some_func(arr, what, ever, sched=None):
    sched.submit(....)
```

dmpelt:
Sounds useful, thanks! We already have quite a few methods that use cf at the moment, so we will have to change it at more places that just this PR.

Would the scheduler typically change between function calls? If not, we could also add something like a `tomopy.set_scheduler(sched)` method that sets a global scheduler that is used in all calls, and default to ThreadPoolExecutor if it is not set.

I'll look into a good way of implementing this once I start working more on this PR - maybe we will have to create a separate PR for this.

EricPythonMaster:
@dmpelt Schedulers can be reused between function calls.

I suggest that calls to CF should be centralized, as it was with the distribute_jobs for multiprocessing.

tomopy/prep/phase.py
```diff
@@ -119,36 +129,47 @@ def retrieve_phase(
...
-            filtproj, planner_effort=_plan_effort(num_jobs))
-            ) / phase_filter.max()
-        if pad:
-            proj = proj[px:dy + px, py:dz + py]
-        tomo[m] = proj
+    prj = pyfftw.n_byte_align_empty((dx, dy + 2 * py, dz + 2 * pz), pyfftw.simd_alignment, dtype='complex64')
+    val = np.complex64(val)
+    e = cf.ThreadPoolExecutor(ncore)
+
+    def setv(a, b, c, l, r, u, d):
```
EricPythonMaster: Instead of using concurrent futures here, have you considered using numba vectorize with a parallel target?
http://numba.pydata.org/numba-doc/latest/user/vectorize.html

carterbox:
Because of #335, the commit hashes for the main branch have changed, you will need to cherry-pick the commits from this pull onto the new tomopy/master.

dgursoy:
This PR seems to be outdated and will be closed for now.

PR #280, opened 2017-09-14
find_center_vo coarse incorporate 4x ds

mdw771:
Changed `smin` and `smax` arguments in call to `_search_coarse` in the large sinograms case to `smin / 4.0` and `smax /  4.0` in order to incorporate downsampling.

### DIFF
--- a/tomopy/recon/rotation.py
+++ b/tomopy/recon/rotation.py
@@ -247,7 +247,7 @@ def find_center_vo(tomo, ind=None, smin=-50, smax=50, srad=6, step=0.5,
     # Coarse and fine searches for finding the rotation center.
     if _tomo.shape[0] * _tomo.shape[1] > 4e6:  # If data is large (>2kx2k)
         _tomo_coarse = downsample(np.expand_dims(_tomo_cs,1), level=2)[:, 0, :]
-        init_cen = _search_coarse(_tomo_coarse, smin, smax, ratio, drop)
+        init_cen = _search_coarse(_tomo_coarse, smin / 4.0, smax / 4.0, ratio, drop)
         fine_cen = _search_fine(_tomo_fs, srad, step, init_cen*4, ratio, drop)
     else:
         init_cen = _search_coarse(_tomo_cs, smin, smax, ratio, drop)


nghia-vo:
It doesn't harm much to search large area (defined by smin, smax). This helps to avoid local minima, and suitable for the case where center of rotation is off-set too much. Another thing: smin, smax have to be integer.  I recommend not to merge this.

mdw771:
I think the issue here is that the actual location of the center can be excluded from the search range in the case that downsampling is applied to the input sinogram. 
In the implementation here, the sinogram will be downsized 4 times if its size is larger than 2k x 2k (at l248):
```
if _tomo.shape[0] * _tomo.shape[1] > 4e6:  # If data is large (>2kx2k)
    tomo_coarse = downsample(np.expand_dims(_tomo_cs,1), level=2)[:, 0, :]
    init_cen = _search_coarse(_tomo_coarse, smin, smax, ratio, drop)
    fine_cen = _search_fine(_tomo_fs, srad, step, init_cen*4, ratio, drop)
```
In this case, if `smin` and `smax` are both negative or both positive, the actual center can be missed if they are not reduced accordingly. 
For example, the original sinogram size is 4000x4000 and the center location is 1500. The specified `smin` and `smax` are -600 and -400. In the first-pass coarse search, the sinogram size is reduced to 1000x1000. Now the center location become 375, which is out of the search range.

nghia-vo:
smax and smin are count from the middle of the image, not from the left or right. For example, if the sinogram width is 2000, the search range for CoR is (2000/2+smin->2000/2+smax). Of course you can change smin and smax from the default values to cover larger range, but not dividing them. In your example, the downsampled image width is 1000, CoR is 375, the default search range is (1000/2-50; 1000/2+50)=(450;550). So you need to extend smin to something like -200. If you do smin/4 and smax/4 -> the search range is (1000/2-50/4; 1000/2+50/4)=(487.5; 512.5) which means that you narrow the search range. In practice (synchrotron beamlines), we often do alignment for the CoR stay around the middle of the image, that why the default values (+/-50) is not much offset from the center of the image. You can increase this range if you would like to.

dgursoy:
@mdw771 @nghia-vo What is the status of this one?

mdw771:
Oops sorry I forgot to follow up. Nghia, I still suggest the proposed change. The problem here is not the width of the search window; instead, the position of the window can be probably shifted in the downsampling case if the user set `smin` and `smax` both negative or both position, instead of being `smin < 0` and `smax > 0`, and this can cause the true center to end up getting missed from the window. The scenario happens when the true center is far from the middle of the image - which can often be the case if the acquisition is not carefully calibrated or if mosaic tomography is conducted. Let me reiterate my example scenario for this circumstance: the original sinogram size is 4000x4000 and the center location is 1500. The specified `smin` and `smax` are -600 and -400. In the coarse search, the sinogram size is reduced to 1000x1000. Now the center location become 375. If the original  `smin` and `smax` of -600 and -400 are used in the downsampled image, then the actual search range is `(1000/2-600, 1000/2-400) = (-100, 100)`. Then the coarse search will not return the correct result at all, because the true answer 375 is out of the window (-100, 100). To fix this, you have to downscale your `smin` and `smax` to -600/4=-150 and -400/4=-100, so that the actual window can be `(1000/2-150, 1000/2-100)=(350, 400)`.

nghia-vo:
From algorithmic point of view, it's correct that the size of the search window should be downsampled as the size of the sinogram. From practical point of view, people doing tomography always try to put rotation axis to the middle of the image (it's not a difficult thing to do), otherwise you gonna lose reconstructable area. That why the reference point of the search window is chosen at the middle of the image. Anyhow, your suggestion is a good point for non-experienced users.

dgursoy:
In very few cases you can't put the rotation center in the middle for imaging samples that are larger than the detector field of view. I think @mdw771 is trying to say this. I am accepting this PR, since it seems to be back-compatible with the normal tomography scans where the rotation axis is nearby the detector's field of view.

PR #287, opened 2017-10-23
ENH: Add use of MKL's FFT when USE_MKL is set

oleksandr-pavlyk:
If `use_mkl` in `setup.py` is set to True, `gridrec.c` is using "mkl.h" and
use Dfti* functions to compute FFT transforms.

This resolves the crash in FFTW resulting from non-thread safety.

@dgursoy


dgursoy:
This fixes #287

PR #296, opened 2017-11-14
Add support for reconstruction with UFO

matze:
This change adds the wrapper which we currently distribute with UFO itself (and would remove if this gets merged). This allows reconstruction with the FBP and DFI algorithms, the iterative methods require a more intricate installation procedure, so for now this is kept outside. The wrapper has no build time dependency but of course requires a functioning UFO installation at run-time.

At the moment there is one caveat which is also documented: the user *must* call TomoPy's reconstruction routine with `ncore=1` otherwise our own scheduler gets confused. Ideally, the reconstruction plugins could either override this decision or at least give this information to the calling function.

### DIFF
--- a/doc/source/ipynb/ufo.rst
+++ b/doc/source/ipynb/ufo.rst
@@ -14,13 +14,8 @@ reconstruction algorithms.
 Install
 `TomoPy <http://tomopy.readthedocs.io/en/latest/install.html>`__,
 `ufo-core <http://ufo-core.readthedocs.io/en/latest/>`__ and
-`ufo-filters <http://ufo-filters.readthedocs.io/en/master/>`__ and
-import TomoPy and the TomoPy wrapper for UFO:
-
-.. code:: python
-
-    import tomopy
-    import ufo.tomopy
+`ufo-filters <http://ufo-filters.readthedocs.io/en/master/>`__. Make sure to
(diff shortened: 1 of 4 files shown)


doc/source/ipynb/ufo.rst
```diff
@@ -108,7 +103,7 @@ than 1 will crash the reconstruction.
 
 .. code:: python
 
-    recon = tomopy.recon(proj, theta, center=center, algorithm=ufo.tomopy.fbp, ncore=1)
+    recon = tomopy.recon(proj, theta, center=center, algorithm=ufo_fbp, ncore=1)
```
dgursoy: @matze Could you please check if the UFO reconstruction example in docs is using the same notation?

dgursoy:
@matze Thanks for this PR! Before closing this, could you please add and answer "What is UFO?" question in FAQ page in docs and explain the basics of its capabilities and usage for general users?

doc/source/ipynb/ufo.rst
matze: This *is* the example in the documentation or do I misunderstand something?

dgursoy: Ah. My confusion. Sorry.

Issue #271, opened 2017-06-29
Problem loading "libtomopy.pyd"


teracamo:
Tomopy works fine on Windows 10 ipython. However, when I try to script with pyCharm and run, the code always goes wrong when loading the library. 

`c_shared_library('libtomopy')` always raise error at `ctypes._dlopen(...)`

I tried on iPython, if I directly load "libtomopy.pyd" with `ctypes._dlopen`, it will go wrong unless I 
`import pyffwt` first, which is quite strange. This trick doesn't work in python scripts. In fact this trick only works in iPython and not even in normal python console.

Would love to get this solved.

teracamo:
I forgot to mention, the error was 
```WindowsError [126]: The specified module could not be found.```
I checked online, and tried to install msvc redist c++, doesn't work.

teracamo:
I tested it on linux platform for my build and loading libtomopy.so with ctypes._dlopen() seems normal.

newville:
@teracamo: these should probably be using `ctypes.windll.LoadLibrary()` on windows and `ctypes.cdll.LoadLibrary()` on linux and darwin.

teracamo:
@newville 

> these should probably be using ctypes.windll.LoadLibrary() on windows and ctypes.cdll.LoadLibrary() on linux and darwin.

I have tested both function on the window machine, both doesn't seems to be working, or are they pointing to the same function for windows as an overload?

PR #299, opened 2017-11-19
Add CI for Windows

thejinx0r:
I made some fixes so that it would compile on Windows with MinGW.

Tests were passing on 3.6 and 3.4.
3.5 was not compiling because conda couldn't find mkl-devel for 3.5, but I had since added the intel channel.

It compiled fine on 2.7, however tests were failing because it couldn't import LIBTOMOPY for some reason.

Tests are now failing because scipy.msic.imresize had been marked deprecated is now removed.
So I can't really progress any more until that is resolved.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -65,8 +65,8 @@ install:
   - set PYTHONHOME=%MINICONDA%
   - conda config --set always_yes yes --set changeps1 no
   - conda config --add channels conda-forge 
-  - conda config --add channels intel 
   - if %PYTHON_MAJOR% == 2 conda install --yes futures nose six numpy h5py scipy scikit-image pywavelets mkl-devel mkl_fft python-coveralls dxchange numexpr python # force update to python >=2.7.14 to avoid  bug with load library introduced in 2.7.13
+  - if %PYTHON_MAJOR% == 3 conda config --add channels intel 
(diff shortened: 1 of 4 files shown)


dgursoy:
@thejinx0r I don't know about why LIBTOMOPY can't be imported in Python 2.7. Does it have a different environment setup, or it can be imported but the builds are not properly done and it complains?

But, now all deprecated `scipy.misc.imresize` functions are replaced with `skimage.transform.resize`. The output is not exactly matching, but we don't have a test for this yet and no functions are dependent on these yet. So it should be fine. After these changes are the tests passing now in Python 3?

thejinx0r:
I rebased my changes on the current master and it compiles and passes for 3.4-3.6, but the tests still fail for 2.7.

It is unable to load libtomopy.pyd in 2.7. It might be related to #271.
I checked the dependencies and they are there. So I'm not sure what's different between 2.7 and 3.x series.

dgursoy:
Thanks @thejinx0r! This is a great! 

Then before I accept this PR, could you please inactivate Python 2.7 builds in AppVeyor? Because I think it may be misleading for developers (the tests are failing not because of the changes in code). We can continue working on the issue in #271 and we activate 2.7 builds once it is fixed.

PR #288, opened 2017-10-23
added filter choice in write_center

vdeandrade:

PR #305, opened 2017-12-07
added filter choice in write_center

decarlof:


decarlof:
@vdeandrade, @dgursoy this replaces [288](https://github.com/tomopy/tomopy/pull/288).

decarlof:
is failing on the test_rotation but the new added parameters are optional (algorithm='gridrec', filter_name='parzen') so the test call:

        write_center(
            read_file('proj.npy'),
            read_file('angle.npy'),
            dpath, cen_range=cen_range)

should still be OK, no?

dgursoy:
I think default was 'shepp', not 'parzen'. This might be the reason that the test is failing. Maybe this would work?
```
write_center(
        read_file('proj.npy'),
        read_file('angle.npy'),
        dpath, cen_range=cen_range, 
        algorithm='gridrec', filter_name='shepp')
```

dgursoy:
'write_center' test is now passing, but 'find_center' test is failing because it is also sharing the same functions underneath, and needs to be updated as well. Once it is fixed I believe this is ready to go.

dgursoy:
All good to go. Thanks!

decarlof:
@dgursoy wait on @vdeandrade comments before accept it. The old pull request from @vdeandrade had changes besides write_center in  _adjust_hist_limits and _adjust_hist_limits (see original pull request at  [288](https://github.com/tomopy/tomopy/pull/288) for details). These are not needed to add the filter choice in write_center so I removed them. @vdeandrade do you remember what other functionality you wanted to add here?

decarlof:
@dgursoy ops too late ....

PR #310, opened 2018-03-28
changes to setup.py for Windows builds 

newville:
With these changes to `setup.py`,  tomopy builds, installs, and runs (at least gridec reconstruction with the "tooth" example from the docs) for Anaconda Python3.6.4 on Windows 10, using mingw64 6.3.0. 

The changes here also comes close to building for Python 2.7, but that requires and additional hack to  `distuils/cygwinccompiler.py` that is not easily monkey-patched, and even then although `recon` runs, the results are not sensible....  It might be that Python 2.7 is not worth trying to support.

I have not yet been able to build a conda package for this.


dgursoy:
Thanks @newville. Any idea why Python 3.5 Linux build is failing after your change?

dgursoy:
Should we keep this PR open?

newville:
Yes, please leave open. I learned today that the trick here of pointing to pkgs is not going to work with conda build.  I will explore building and including a precompiled dll for gridrec instead of trying to build that at install time. 
That's what I do for pyepics and Larch.

carterbox:
Because of #335, the commit hashes for the main branch have changed, you will need to cherry-pick the commits from this pull onto the new tomopy/master.

newville:
@carterbox @dgursoy in fact, I will close this PR and attempt another approach, as mentioned briefly last week, and I'll expand on here:

That is, this PR *can* work to build on Windows, but relies on a compilation hack of pointing to `pkgs` for Windows library files that are needed for building but not for running code.  This approach does then work when a) the correct mingw compiler is installed and in the PATH, and b) the mkl-devel and mkl-include packages are both installed and placed in a findable `pkgs` directory.  This is *not* the case for a build with conda (that creates its own build enviro, and so without `pkgs`, or at least with `pkgs` in a not-easy-to-find place).

But this then leads to a larger question of "Why does tomopy compile extensions at install time"?  This might be understandable if the idea is that every environment has a full and working compiler set and/or that libraries built on one machine may not be easily used on another machine.   But, that really only describes some Linux systems.  For sure, libraries *are* portable for a particular architecture on MacOS and 64-bit Windows.

So, my plan is to precompile shared libraries for Windows and add a `setup_win.py` that does not try to compile `tomopy lib` and instead load it from a precompiled version included with tomopy.  Please let me know if you have any concerns with this approach.

carterbox:
Wait, what!? I thought that the point of `conda` was that packages were precompiled for every target architecture? To clarify, is the problem that the server where the distributed packages are compiled doesn't have the necessary tools to compile for Windows?

newville:
@cartedbox

> Wait, what!? I thought that the point of conda was that packages were precompiled for every 
> target architecture? To clarify, is the problem that the server where the distributed packages 
> are compiled doesn't have the necessary tools to compile for Windows?

A conda package is distributed as binary, but is *built* at `conda build time` which runs `python setup.py install` in a special, temporary environment.   The way tomopy's `setup.py` is written, `python setup.py build` will *always* compile libtomopy.cpython-***.so .   

On Windows, the compilation of libtomopy  is complicated by the facts that `gridrec` needs a C99 compiler -- so *not* the default MSVC --  because it was rewritten (a couple years back) to use `<Complex.h>`. In addition, `gridrec` needs to link with mkl runtime.  On Windows this need to find the `mkl_rt.lib` file which is not installed, but can be found in the `pkgs` folder for `mkl-2018` or something.  So, it *can* be hacked to work with a different compiler and point to the `pkgs` folder, but the "pointing to `pkgs`" hack does not work (or I could not figure it out)  within the environment used to build the conda package.   Then again, if you can build a conda package for Windows, go for it.

Anyway, why is libtomopy.cpython-***.so built at `python setup.py build` time at all?   For Windows 64 (and MacOS for that matter) it should be OK to build it exactly once ever on one machine (well, any time the C code changes), not at `python setup.py build` time.

carterbox:
I support this approach because it makes sense in order to make the Windows version distributable. However, for transparency, we will need to have detailed directions for building the Windows library, and for security we will have to disallow any pulls with changes to the Windows library from strangers.

PR #324, opened 2018-05-30
Fixing bug in gridrec.ipynb

kmader:
`dxchange.read_aps_32id` outputs 4 arguments not 3 and so added theta to output (which later gets overwritten)


dgursoy:
Good catch. Thanks.

PR #325, opened 2018-05-30
Adding Binder links and demo notebook

kmader:
With binder you can make interactive demos of tomopy that run in the browser so people can try it without installing anything. I've added links for one or two notebooks, but more examples would be good.


dgursoy:
This is a great feature to have. Thanks. I think we can/should now put more examples there.

Issue #301, opened 2017-11-19
Windows 10 - distribute wheels


thejinx0r:
This should be more of a discussion than a real issue.

I added tomopy to AppVeyor which is the Windows CI ( #299 ).
Assuming ( #300 ) gets resolved quickly, we should be able to generate python wheels for Windows users.

My question / discussion point is: how should the wheels be distributed?

- The precompiled binaries can't be distributed through conda-forge since it doesn't support gcc for windows (aka mingw or msys2).

- We could do releases on PyPI, but I have no experience with it and I don't know if it's a good idea or not.

This issue relates to:
#283 #271

dgursoy:
Is there a workaround for the first item? There is a long thread here about a similar subject:
https://github.com/conda-forge/conda-forge.github.io/issues/112

We can set up a separate distribution just for Windows (first I can think of is PyPI), but I don't think it will be easier to support in the long run. I'd really want to see if we can find a way to make it conda-forge compatible.

thejinx0r:
It could be possible.
I can check it out in the new year. I don't think I will have time to flesh this idea out until then.

jasonthiese:
Do you still plan to make a tomopy wheel for Windows?  I have tried installing from source, so far without success, in both python 2 and 3.  The Visual C++ compiler that works with python 2 apparently doesn't support C99, so that failed.  (The exact error message:  "...\tomopy\src\utils.h(50) : fatal error C1083: Cannot open include file: 'stdbool.h': No such file or directory".)  I'm still waiting for the compiler that works with python 3 to be installed -- where I work, one cannot simply install software oneself.

thejinx0r:
I'm pretty sure there is a way to download the binaries from the ci builds.

I accidentally wiped my windows partition on my laptop and can't test the
build system anymore.

PR #327, opened 2018-06-06
add gcc-libs as a requirement to build and run on windows

thejinx0r:
As per #301, I tried adding the m2w64-gcc-libs and it's working.
I'm creating a pull request also to trigger travis to make sure that the changes to the meta.yml won't cause build issues on linux.

The only issue I encountered so far is that on the users's machine, I needed to edit the path environment variable to use tomopy properly.

### DIFF
--- a/meta.yaml
+++ b/meta.yaml
@@ -14,6 +14,7 @@ requirements:
     - mkl-devel
     - python
     - setuptools
+    - m2w64-gcc # [win]
 
   run:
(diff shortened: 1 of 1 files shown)


thejinx0r:
Hopefully, this means that they will be available on windows too through conda-forge now.
I tried a conda-build on my windows work machine, but it has built it in the past.

PR #332, opened 2018-08-23
Update build instructions and setup.py

### LINKED ISSUE #174: improve doc for developers
Because we now have some external packages, the installation is not trivial. We need to improve doc for developers explaining how non-conda build will be.

carterbox:
Updated documentation and setup.py to make building and linking with conda installed mkl libraries easier. Closes #174


dgursoy:
Thanks.

PR #333, opened 2018-08-28
Cache intersections in iterative methods

carterbox:
Instead of computing intersections between each ray and the grid as needed, compute intersections for all arrays in advance then use a lookup table to quickly find intersections. This dramatically speeds up these functions because (1) intersections can be computed in parallel and (2) intersections do not need to be recomputed for every iteration. The larger the number of iterations, the more efficient this method becomes. This commit increases the memory footprint in exchange for speed. The increase in the size of the footprint depends on the input data.

I did not study whether the increased memory footprint is problematic for users with low-memory workstations because I don't know what is considered low-memory or what the size of a typical data set is these days.


prjemian:
+1

carterbox:
This pull may contain a bug in which `simdata` is always calculated from the 0th slice. I didn't test on multiple slice reconstructions.

dgursoy:
Thanks @carterbox. Then let's wait for you to address the 0th slice issue before merging it with the main branch.

carterbox:
I fixed that bug. It's the commit after I mentioned the bug.

carterbox:
The changes in this pull should be tested with a real sized data set so we can get an idea of the memory requirements before it's merged. @prjemian, could you run a test for me?

prjemian:
@carterbox : my system is not ready to test tomopy now, let's ask someone with more experience running tomopy.  Perhaps @dgursoy ?  @decarlof ?

dgursoy:
Could you resolve the conflict before the merge?

carterbox:
I finally got around to testing this feature on a 1501 angle, 1 slice, 2048 wide dataset. It's not good news. Caching all of the intersections takes more than 70GiB of RAM before `compute_indices_and_lengths` segfaults.

Back of the hand calculation for the number of intersections **and** lengths that need to be stored:
```
1501 angles x 2048 rays x (2048 x 2) = 12,591,300,608
```
C99 data type maximum values:
```
16bit unsigned <= 65,535
32bit long <= 2,147,483,647
32bit unsigned long <= 4,294,967,295
64bit long long <= 9,223,372,036,854,775,807
```
Thus, the RAM required to reconstruct this slice is approximately:
```
12,591,300,608 x (64 bit indices + 32 bit lengths) = 140.71 GiB
```
The main reason this footprint is so unreasonably large is that we needed to use 64 bit types to store the intersections if we use a single buffer. For only 32 bit types, the RAM required is much less:
``` 
1504 x 2048^2 * 2 * (32 + 32) =  94.00 GiB
1504 x 1024^2 * 2 * (32 + 32) =  23.50 GiB
1504 x 512^2 * 2 * (32 + 32) =  5.88 GiB
```
I need to come up with a more sparse representation for this intersection information.

dgursoy:
This is probably the best option to speed things up if you split the work to multiple computers, because you reduce the memory requirement per node, but agree that it may not be a good option for reconstructing mu-CT data on workstations. 

How about you hold only some of these arrays on memory depending on the RAM you have? The more memory you have, the faster it will be then.

carterbox:
By avoiding 64 bit types and using some 16 bit types I estimate I can get the memory footprint down to around 64 GiB for 1501 angles and 2048 wide, but that's still a lot more than what a typical home computer has (3GiB - 8GiB). This is the intersections cache only! I'm not even counting the space needed to store the reconstructed or input data or run other programs.

What is the amount of available RAM that our target user has available!?

If the intersections buffer has a size depending on what is available on the system, then which intersections do you keep? Are some intersections more expensive to compute than others? If we were ray tracing, then I would say it is the ones that have the longest run across the reconstruction grid, but with Siddon's method, they all cost the same?

dgursoy:
I think it’s very challenging to have a fast  implementation of the iterative algorithms for workstations of any size (minutes for mu-CT data). For these types of data, I’d either use a cluster, or GPU kernels for fast forward and backward operations implemented using FFTs. We don’t support the former, and have ASTRA et al. for the latter.

carterbox:
The updated comparison for the single slice reconstruction:

Method | RAM utilization | Time
--|--|--
cached | 57 651 MiB | 147 s
v1.1.4| 00 150 MiB | 171 s

carterbox:
I'm closing this pull request because this feature clearly needs rethought and a lot of work in order for it to be released.

PR #334, opened 2018-09-12
Add Total Variation reconstruction method and Gradient descent method

math-vrn:
Main differences
1) tv.c file was written according to the syntax of other sources for the iterative reconstruction like mlem.c, art.c, etc.
2) reference to the tv method was added in doc/souce/bibtex/refs.bib file
3) tv.npy file was added for performing unit tests
4) grad.c file was written according to the syntax of other sources for the iterative reconstruction like mlem.c, art.c, etc.
5) grad.npy file was added for performing unit tests

PR #336, opened 2018-09-17
 Total Variation reconstruction method and Gradient descent method

math-vrn:


dgursoy:
This is a version of #334 with a cleaned PR history.

PR #338, opened 2018-09-18
Fix segmentation fault in gridrec.c for the non-mkl version.  test_rotation.py passed after fixing this problem

math-vrn:
Segmentation fault appeared in Lines 436,437
p = P_z[z];
j = J_z[z];
because zlimit is sometimes bigger than the number of initialized members of P_z and J_z.

test_rotation.py test passed after fixing this problem

### DIFF
--- a/src/grad.c
+++ b/src/grad.c
@@ -44,213 +44,213 @@
 #include "utils.h"
 
 
-	void 
+    void 
 grad(
-		const float *data, int dy, int dt, int dx,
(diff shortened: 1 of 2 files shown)


math-vrn:
In the gradient-descent method computation of the gradient step should be independent for different slices, otherwise the multi-core version gives different result with the one-core version. Fixed in grad.c

dgursoy:
Related to #336

PR #339, opened 2018-09-18
allow build precompiled libtomopy outside of setup.py, better support for loading Windows shared library

newville:
This PR provides makefiles and simply python configuration script to support building the shared library `libtomopy` outside of `python setup.py build_ext`.  It adds the following file and mechanism:

-   `setup_nocompile.py`:  build tomopy *without* trying to build the shared library for the extensions.  This assumes that a precompiled shared lib will be found.

- `tomopy/sharedlibs/libtomopy.dll` and `tomopy/sharedlibs/libtomopy.dylib`: precompiled libraries for Windows64 and MacOS.

- `src/configure.py`, `src/Mk.base`, `src/Makefile.linux`, `src/Makefile.darwin`, `src\Makefile.windows`: scripts to build and install precompiled shared libraries.   The build process  using these scripts is:
 ```
        python configure.py 
        make install
```

These Makefiles for libtomopy are really not very complicated, but may need some maintenance as compilers and toolkits evolve.

To accomodate the precompiled shared libraries, a few changes were made to  the `c_shared_libs()` function in `tomopy/util/extern.py`.  This now looks first for the "normal" installation of `libtomopy` shared library as built by `setup.py build_ext`.  If that is not found (say, because `setup_nocompile.py` was used), then it will look for the OS-specific precompiled shared library in `tomopy/sharedlibs`.

I have a conda recipe that uses `setup_nocompile.py` for Windows and MacOS (see https://github.com/xraypy/conda_recipes/tree/master/tomopy) and have used this to build conda packages for MacOS and Windows with the precompiled shared libraries.   These conda packages are at https://anaconda.org/GSECARS/tomopy

### CODE REVIEW

**Comment by tacaswell** [comment]
Please do not do this.  It is a mistake on pyepics and will be a mistake here.  

If you want to distribute non-conda binaries, make wheels.

Messing with the library search path is also a bad idea.

**Comment by newville** [comment]
@tacaswell  Sorry, but what exactly is the mistake being made?  I don't understand the criticism of pyepics here either.  What is the objection?  

I would be very happy to have anyone else provide conda packages for tomopy on Window64 or provide a better way to build tomopy on Windows.  I am not aware of anyone offering a conda package for Windows, and several  github Issues point to the challenges of even building tomopy for Windows. 

If you have a better approach, by all means share it.

**Comment by tacaswell** [comment]
As a general rule, built libraries should never be checked into git.  It bloats the git repo and in general will not actually work.

You should also never mess with the library search path at the library level.  While it make locally work, if everyone does it, it leads to chaos (and in fact we have ad issues with pyepics doing this).

In general, you should not solve things that are fundamentally packaging problems at the source level.

For linux follow the manylinux wheels instructions: https://github.com/pypa/manylinux

For mac and windows I think you can follow https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives directly.

For windows conda packages work with the conda-forge people.  It sounds like you have successfully built windows conda packages so why do we need to push the binaries into the source repo?

As for pyepics, I think including libca and friends in the git repo is a big mistake, but settled for ensuring that there is way to make sure they do not get installed (rather than deleting them from git).

**Comment by newville** [comment]
On Tue, Sep 18, 2018 at 1:27 PM Thomas A Caswell <notifications@github.com>
wrote:

> As a general rule, built libraries should never be checked into git. It
> bloats the git repo and in general will not actually work.
>
Hm, appealling to "as a general rules" is pretty vague. Do you have
examples of actual problems with this PR?

As for bloat,  libtomopy.dll is 100 kb.   Since I also added
libtomopy.dylib, the total added bytes is ~250 Kb, about the size all the
src/*.c files.  More than 100 mb of data was removed from the repository in
the past week.   As it turns out, I was the one who complained about that.
Apparently a few 100 kb doesn't really qualify as bloat.

Most importantly, you say that precompiled libraries will not actually
work.  Do you have any evidence for this? They work for me all the time,
and this one works for me here, at least in the few tests I ran.

You should also never mess with the library search path at the library
> level. While it make locally work, if everyone does it, it leads to chaos
> (and in fact we have ad issues with pyepics doing this).
>

This PR does not "mess with the search path" for shared libraries any more
than the tomopy master branch does.  In the current master, the code to
find the shared library is

     if os.name == 'nt':

            ext = '.pyd'

        else:

            ext = '.so'

        _fname = os.path.abspath(os.path.join(os.path.dirname(__file__),
'..'))

        libpath = glob.glob(_fname + '/' + lib_name + '*' + ext)[0]

        return ctypes.CDLL(libpath)

    except (OSError, IndexError):

        logger.warning('OSError: Shared library missing.')



This relies on the behavior of `setup.py build_ext` to place a
properly-named shared library in the `tomopy` directory at build time.  I
suspect the use of `glob` here is because although there will be exactly
one shared library, the actual name of the file is hard to accurately
predict.

This PR adds a test for a specific file in the `tomopy/sharedlibs` folder
based on OS. That test only runs if the above search for a shared library
in the `tomopy` folder has failed, as when the extension is not built with
`setup.py`.  The code added in this PR is:

    # alternate -- shared lib built by 'make install' in src directory

    ext = '.so'

    if sys.platform == 'darwin':

        ext = '.dylib'

    if os.name == 'nt':

        ext = '.dll'

        load_dll = ctypes.windll.LoadLibrary

    sharedlib = os.path.join(_fname, 'sharedlibs', '%s%s' % (lib_name, ext))

    if os.path.exists(sharedlib):

        return load_dll(sharedlib)



If `os.path.exists()` is messing with search paths then I plead guilty. But
I also wonder why you fnd `os.path.exists()` to be "messing with search
paths" but the use of `glob()` to raise no concerns.

With this PR,  an installation done with `python setup.py install` will
build, install, and find the shared library exactly as before, and the
newly added code should not ever be run unless the installation is done
with `python setup_nocompile.py install`.

I see no chaos here.  There is either a locally built shared library or a
precompiled library that is used.  It does nor interfere with other
packages.  I don't know what chaos you are referring to that pyepics causes.

In general, you should not solve things that are fundamentally packaging
> problems at the source level.
>

This is fundamentally a build problem, not a packaging problem.  Compiling
libtomopy requires a C99 compliant compiler which is not standard for
Windows.

For linux follow the manylinux wheels instructions:
> https://github.com/pypa/manylinux
>

While I did include Makefile scripts to build a shared library for Linux,
and tested them on one Linux box, I did not include a shared library for
Linux.  Building tomopy on Linux seems to be no problem.  MacOS and Windows
shared libraries are remarkably portable.  But I agree with you that if one
wanted to include a libtomopy.so for Linux, building that with manylinux
would be one way to go (then again, CentOS 5 is no longer supported, and
using CentOS 6 would also be defensible).

For mac and windows I think you can follow
> https://packaging.python.org/tutorials/packaging-projects/#generating-distribution-archives
> directly.
>

The problem is not related to installation or uploading packages.  As far
as I know, tomopy is not available on PyPI and no wheels exist for it.
Anyway, I *want* conda packages of tomopy for Windows.

When building a conda package for the tomopy master branch, it will always
compile and link the shared library inside the temporary conda package
environment.  The fundamental problem here is that this linking in the
temporary environment does not work on Windows.  I have tried many
different approaches,  and I know others have too.  They are all very
fragile hacks, and I cannot get them to work with `conda build`. If you or
someone has detailed, specific suggestions or can figure out how to make
that work on Windows, that would be great.  But you do not need to point me
to tutorials on packaging and distribution. ;).

I am definitely not eager to jump into working on yet another python
package, and given the historical hostility of tomopy to Windows and the
initial reaction to this PR, I am definitely not feeling like being more
involved with tomopy would be a good use of my time. But I need tomopy to
work on Windows.  No one else has done this or offered any viable approach,
and the push back here is more that a little alarming.  Honestly, it's kind
of ridiculuous that tomopy is not supported on Windows, and ironic that the
issue is with gridrec, as the original gridrec code from BNL  and altered
by Mark Rivers before being used in tompopy originally worked on Windows.
Gridrec was broken in late 2015 exactly by the combination of following
"best practices" and ignoring Wiindows.   That is the mess I am trying to
work around.

For windows conda packages work with the conda-forge people. It sounds like
> you have successfully built windows conda packages so why do we need to
> push the binaries into the source repo?
>

I don't follow you.  The current master branch will always try to build the
shared library at build_ext time.  And that build will fail on Windows. The
conda package built using this PR relies on the shared library being
installed to site-package/tomopy/sharedlibs without compiling during `conda
build`.  Maybe I'm not understanding your point, or maybe you are not
familar with the actual issues.

As for pyepics, I think including libca and friends in the git repo is a
> big mistake, but settled for ensuring that there is way to make sure they
> do not get installed (rather than deleting them from git).
>

I think you do not understand how pyepics works or why the shared libraries
are included.  It is for sure not a mistake to include the shared libraries
with pyepics.

All that said, the tomopy developers can accept this PR and the gift (or
burden) of code, or not.  Tomopy has not supported Windows in the past.
This PR makes it possible for them to support Windows users going forward.
It also provides an opportunity for them to clearly state that they do not
support Windows.

Either way, I can now build and distribute tomopy for Windows using conda
for my downstream packages and users. The question is really whether the
fork is permanent or gets merged into the main tomopy branch.  That choice
isn't mine.   But I do have to admit that it will be challenging for me to
forget this interaction.

**Comment by tacaswell** [comment]
I am sorry this has gotten personal, however I stand by my technical judgement that it is a bad idea to check build products / binaries into the source repository.

> But I do have to admit that it will be challenging for me to forget this interaction.

@newville please remember your problem is with me, not with the tomopy developers at large.

**Comment by carterbox** [comment]
@newville, I'm sorry that reaction to this pull was not what you hoped for. I realize that you have spent a lot of hours trying to make `tomopy` support Windows, and I thank you for your patience while we try to find the best solution to this problem.

Adding the binaries to the repository will be a high maintenance solution because they have to be recompiled every time that someone makes changes to the source. Who will be responsible for that? We don't have anything against Windows, it's just that none of the main developers of `tomopy` use Windows, so bootstrapping Windows support requires man-hours that haven't been allocated.

I think what @tacaswell is suggesting is that we could deliver the binaries through `conda-forge` using Appveyor. Appveyor is a continuous integration service with a Windows build environment. I didn't know it existed until today, but apparently it has been mentioned before by @thejinx0r in #301 where they were requesting a wheel. In fact, here's a [blog post](https://tschoonj.github.io/blog/2017/10/01/conda-forge-creating-windows-compatible-recipes-for-autotools-based-python-projects/) from the developer of xraylib describing how he used Appveyor to add Windows to his list of target systems. Our current continuous integration service is TravisCI, but it only has a Linux build environment.

If it is possible to build `tomopy` on Appveyor with their list of [preinstalled software](https://www.appveyor.com/docs/windows-images-software/) and @dgursoy and @decarlof approve (they control the `conda` and TravisCI accounts), then it would seem that the best solution is to add or switch to Appveyor as our continuous integration service because then the Windows builds would happen automatically and be deliverable through `conda-forge` just like our Linux builds. (I don't know about MacOS, I wasn't aware that there was a difference with the Linux build.)

File: src/configure.py [hunk 1]
```diff
@@ -65,7 +68,14 @@ def config_windows():
     """ config for Windows"""
     """ config for Linux"""
```
thejinx0r: it should only have config for windows

thejinx0r:
I got mentioned. `
I've never used conda distribute packages and I've never looked in to starting a channel, but last time I tried, doing a git clone followed by "conda build" in the cloned folder was sufficient to build it.
I assume from there, it should be straight forward to distribute the package through conda, but I haven't had the time to look in to it.

newville:
With the latest fixes, this now works to build libtomopy shared libraries on Windows, MacOSX, and Linux. For Windows and Anaconda Python, the MinGW and Make packages are required.

newville:
@carterbox  Thanks!  I have no objection to not storing the shared libs in the source repo if these and binary packages are automatically created by some outside build system.

PR #342, opened 2018-09-30
Wrapper for the Log-polar based method on GPU

math-vrn:
A wrapper for the log-polar based reconstruction method was added to tomopy/recon/wrappers.py, so as simple python functions for reconstructions with the FBP method, gradient descent method, and tv regularization.
Other methods can be added by writing corresponding python functions in wrappers.py
The log-polar based method can be installed from https://github.com/math-vrn/lprec.
For examples see Section 'Docs/Examples/Log-polar based reconstruction' in Tomopy readthedocs page.


math-vrn:
@carterbox , could you help to go through these checking tests. The last error I saw is that mkl_rt library is not found. As I understand, numpy works with this library so the tests should be ok

carterbox:
@math-vrn The Windows tests are failing because you're pulling onto `master` which doesn't have windows support yet. When #344 is merged, then all the tests on this pull will probably pass.

dgursoy:
How can you trigger Appveyor after merge of #344?

carterbox:
My bad. I had set Appveyor to not build the master branch while I was debugging the build settings.

dgursoy:
Thanks @math-vrn!

PR #349, opened 2018-10-13
Move lprec iterative schemes from wrappers.py to the lprec package 

math-vrn:
The lprec package (https://github.com/math-vrn/lprec) now includes iterative schemes: gradient-descent, CG, EM, TV.
Linear algebra operations inside these scheme are accelerated by using cupy module (https://cupy.chainer.org/). Access to GPU data by lprec library works via GPU memory pointers.


dgursoy:
@carterbox

carterbox:
I have no idea what this is about because I'm not familiar with `tomopy.recon.wrappers`.

@math-vrn Added some new feature but then y'all decided it would be better as a separate project? Is `lprec` an optional dependency now?

math-vrn:
@carterbox tomopy.recon.wrappers includes wrappers for reconstructions from different packages: Astra toolbox, UFO, vectorised, lprec. How to use these methods inside tomopy is described here: https://tomopy.readthedocs.io/en/latest/demo.html. As I understand, @dgursoy is asking you to test the lprec method and report any bugs. Let me also know if the documentation is not clear or there are problems with installation.

PR #350, opened 2018-10-15
Change types of default parameters for lprec method (strings-constant compatibility for Python3)

math-vrn:


carterbox:
There are no tests in `tomopy` for the wrapper module. Have you run tests for these wrappers for both python 2 and 3?

dgursoy:
It's a good point, but broadly speaking I think it would be hard to test all wrapped up packages. Because not all external functions are duplicates of internal ones (with tests available) and it is up to the developers of the external package to write their own tests for their software. Besides, their installations can be tricky, requires a specific system, package version, etc.

Issue #351, opened 2018-10-27
align_seq needs center shift control?


jefferton-sg:
Hey all, long time fan, first time caller...

I've been tinkering with the align_seq routine for projection image alignment but I think it needs a control for the center shift. While I've "fixed" this on my end by adding another call to find_center inside the align_seq code, I suppose it would make more sense to have center shift as an optional argument when calling align_seq (so the reconstruction procedure would then only need to do that once). Of course if drifts are REALLY bad, then the center shift finding might need to be done with every iteration of align_seq... in that case, the right place would be inside align_seq itself. 

Any thoughts?

carterbox:
Thanks for you idea! It looks like @dgursoy is the last person to have touched [`align_seq`](https://github.com/tomopy/tomopy/blob/788aaf43ff3db9392629ad0c10f4b29db31063f6/tomopy/prep/alignment.py#L75-L166). He is at a conference this week, but I'll be sure to remind him in person to address this issue when he returns. Thanks for your patience. :smile:

PR #352, opened 2018-10-30
ENH: more args are available in align_seq and align_joint

dgursoy:
This PR addresses #351. 

I added the the following arguments as optional in functions:
- ```center```: The rotation center to be used. 
- ```algorithm```: One can pick the algorithm to be used in the alignment process.
- ```rin``` and ```rout```: These are to define the blur kernel to smooth the projections. These are now set to 0.5 and 0.8 by default.
- ```upsample_factor```: This is used to define the level of accuracy in registration. 10 is set to default, which means 1/10th of a pixel accuracy.

I haven't add automated centering available, because one can use it before calling the alignment function, and provide the resulting center as input. 

Please check and let me know if you have any concerns or feedback. Also see you soon at the workshop @jefferton-sg!

### DIFF
--- a/tomopy/prep/alignment.py
+++ b/tomopy/prep/alignment.py
@@ -74,7 +74,9 @@ __all__ = ['align_seq',
 
 def align_seq(
         prj, ang, fdir='.', iters=10, pad=(0, 0),
-        blur=True, save=False, debug=True):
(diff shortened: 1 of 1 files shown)


dgursoy:
I’ll merge this PR and open a new PR for #358, because I’m not sure if it’s trivial to implement.

PR #353, opened 2018-10-31
Wrapper for lprec multigpu version

math-vrn:
This new wrapper for the log-polar based reconstruction splits the whole number of slices by parts in order to fit them in gpu memory. The parts are dynamically sent to gpus for reconstruction. In this case, faster gpus process a bigger number of slices parts than slower gpus. Users can specify gpu_list variable to define which gpu numbers should be used for reconstruction.

### DIFF
--- a/tomopy/recon/wrappers.py
+++ b/tomopy/recon/wrappers.py
@@ -80,6 +80,7 @@ default_options = {
         'filter_name': 'None',
         'num_iter': 1,
         'reg_par': 1,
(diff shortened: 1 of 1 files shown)


dgursoy:
Thanks @math-vrn! Could you do a PEP check before the merge please?

math-vrn:
I did it with autopep8

dgursoy:
Thanks!

PR #355, opened 2018-11-02
Benchmarking with PyCTest + GCov

jrmadsen:
- `setup.py` imports and runs `build.py` so that the build/installation can be done with `python setup.py install`
- Added `pyctest_tomopy.py` for submitting build, testing, coverage, and benchmarking to CDash
- Added benchmarking folder for various tests
  - `benchmarking/pyctest_phantom.py` - script for testing built-in phantoms
  - `benchmarking/pyctest_rec.py` - script for testing TomoBank data
- Added `tomopy.misc.benchmarking` for routines related to combining, comparing, and joining results
- Integrated PyCTest into Travis
- Enabled setting `CFLAGS` and `LDFLAGS` in Makefile.{darwin,linux} for GCOV coverage
  - GCov coverage is enabled with `--coverage` option to `./pyctest_tomopy.py`


setup.py
```diff
@@ -1,8 +1,96 @@
...
+        print("")
+        try:
+            import shutil
+            shutil.copy2(src, dst)
+        except Exception as e:
+            print("Exception occurred copying '{}' to '{}'... {}".format(src, dst, e))
+
+
+# ---------------------------------------------------------------------------- #
+#
```
newville: Minor point: why not use `os.path.join(os.getcwd(), "tomopy", "sharedlibs", libname)` instead of your `join_path_list`?

Larger point:  All this code is added to replace `python build.py` as an initial step.   If that is  actually needed,  maybe consider something like `subprocess.check_call(('python', 'build.py'))`.  That would be a lot less new code.

carterbox: About the larger point, can you explain why this addition is needed instead of just calling `build.py` from inside `setup.py`?

jrmadsen: @newville w.r.t. minor point -- I'll fix that. Used to more C/C++ so when it didn't accept a list, I just assumed I'd have to do that

jrmadsen: w.r.t. larger point

- I figured this style was the better option since it gives access to the standard setup.py installation environment
- I can modify it to that scheme if you want but, in the end, I might not matter if y'all want to integrate a CMake build system (I have this done elsewhere)
- CMake will enable a number of improvements over the current `build.py` system, here are a few off the top of my head:  
    - Improved platform portability for compiled code
    - Improve compiler support  (MSVC, Clang, PGI, etc.)
    - More customizable build options without dependence on knowing specific compiler flags
    - Better package support configuration (e.g. enable MPI, CUDA, OpenMP, etc.) under-the-hood for compiled code
    - Easily enable C++ support
    - It will be much easier to generate code coverage for C/C++ code
    - Ability to run clang-tidy on source code

setup.py
```diff
@@ -1,8 +1,96 @@
...
+    else:
+        path = join_path_list([os.getcwd(), "tomopy", "sharedlibs", "libtomopy"])
+        for ext in ["so", "dylib", "dll"]:
+            tmp = "{}.{}".format(path, ext)
+            if os.path.exists(tmp):
+                return tmp
+
+    # return best guess
+    return join_path_list([os.getcwd(), "tomopy", "sharedlibs", libname])
+
```
carterbox: This function is very confusing to me. I believe this function is supposed to return the location of the shared library named `libtomopy.*` whether or not it exists. Why does it check for existence if its just going to return a non-existing file anyway?

setup.py
```diff
@@ -1,8 +1,96 @@
...
+    libname = None
+    if sys.platform.lower().startswith('win'):
+        libname = 'libtomopy.dll'
+    elif sys.platform == 'darwin':
+        libname = 'libtomopy.dylib'
+    else:
+        libname = 'libtomopy.so'
+
+    path = join_path_list([os.getcwd(), "tomopy", "sharedlibs", libname])
```
carterbox: Instead of using all these cases, you could use `glob` and a regular expression.
```python
glob.glob('./tomopy/sharedlibs/libtomopy.*')
```

setup.py
jrmadsen: C/C++ habit of always returning the same type -- I return a string because a string is expected. The try-catch block around `shutil.copy2(src, dst)` is missing a re-raise after it printed diagnostic info on where it was looking for the files and folder it was trying to copy to

jrmadsen: Also, this method covers the three primary errors that will occur in one check: 

  1) file doesn't exist because build failed
  2) insufficient permissions issue causing the copy to fail
  3) the target directory doesn't exist. 

This style enables all of those to be relegated to a single try-catch block without the need for a `if src is not None:` along with the check that the copy succeeded

setup.py
jrmadsen: That could work for the second part. I think the first check would be recommended to be certain that you give priority to the platform specific library since this could cause problems with Docker -- e.g. `docker run -it --rm -v ${PWD}:/home/tomopy tomopy-dev ./pyctest_tomopy.py` -- would return `libtomopy.dylib` on my Mac if that still existed, which is entirely possible because there isn't an easy `make clean` available.

setup.py
carterbox: @jrmadsen Using CMake instead of Make sounds like it has a lot of benefits, but I believe that switch is independent of the changes that you have made to `setup.py`. You have added a bunch of classes to in order to use function calls intended for the `setuptools.Extension` module, but we don't have a Python extension. We are using `ctypes` to load an external library. Could you please explain benefit of telling `setuptools` that we have an extension when we don't actually have an extension?

jrmadsen: I am pretty sure a Python extension is just defined as any compiled code with Python wrappers. So I used `setuptools.Extension` because thats how you create a Python compiled extension.

setup.py
carterbox: I think a more pythonic way would have been to return `None` if the shared library doesn't exist because trying to copy `None` also raises and exception at `shutil.copy2` that you can catch.

jrmadsen: Then we end up with:
```python
        try:
            src_base = os.path.basename(src)
            dst = os.path.join(extdir, "tomopy", "sharedlibs", src_base)
            shutil.copy2(src, dst)
        except Exception as e:
            print("Exception occurred copying '{}' to '{}'...".format(src, dst))
            raise e
```
now neither `src` nor `dst` will possibly be set so the error message is much less informative:
```
"Exception occurred copying 'None' to 'None'..."
```

jrmadsen: And since a failure in `src_base` occurs before `dst` is assigned or even declared, I believe we would then raise an exception within an exception

jrmadsen: I care less about making things "Pythonic" than I do about making sure that the code is safe and when it does fail, provides actual information about what went wrong instead of just a backtrace to where the error occurred.

setup.py
carterbox: Good point. I didn't consider that you would be building for multiple platforms on one machine.

setup.py
newville: @jrmadsen 

> w.r.t. larger point
> I figured this style was the better option since it gives access to the standard setup.py 
> installation environment I can modify it to that scheme if you want but, in the end, I 
> might not matter if y'all want to integrate a CMake build system (I have this done elsewhere)
> CMake will enable a number of improvements over the current build.py system, 

The arguments you make here are advocating for CMake.  I think the concern with the code added to `setup.py` is that it uses way more than 10 lines of code and many levels of indent (that is, complexity) to replace `python build.py` or something like:

```python
    # add this to setup.py?
    from build import buildtomopy
    os.chdir('config')
    build_libtomopy()
    os.chdir('..')
```

That seems completely independent from whether CMake is used or not.

> I am pretty sure a Python extension is just defined as any compiled code with Python 
> wrappers. So I used setuptools.Extension because thats how you create a Python 
> compiled extension.

Well, sort of.  Setuptools seems to expect that an extensions includes python calls in C (or similar) code.  Setuptools will compile the Python+C extensions, often with a `pyd` extension.  That is appropriate for compiled code that really does "<include Python.h>".  When using Cython or ctypes, the definition of extension get blurry.  

It turns out that the compiled code in tomopy does not use Python.h at all and compiles straight C code to a shared library that could be used by (in principle) any language, and is used in tomopy by importing that shared library with `ctypes`.  In that sense,  libtomopy is as much a Python extension as libc is.

So, you can consider libtomopy an extension.  OTOH, expecting that setuptools will automagically compile the code has proven pretty hard, especially on Windows.  If you do want to alter the scripts to build `libtomopy` to build with CMake or to use setuptools.Extension, please make sure the resulting library compiles and works on Windows.

carterbox:
@jrmadsen, Thank you for submitting this pull request. Here's a review of today's discussion about it:

## Benchmarking
I think we're happy to accept these additions of benchmarking for HPC. I assume someone (@dgursoy) reached out to you about implementing them. Otherwise, thanks! :smile: I think tomopy is overdue for some performance optimizations.

## The use of `#---#`.
I can imagine how `#---#` might help break up really long functions in C/C++. In Python, the indentation required by the language and syntax highlighting already do a really great job of making breaks between functions readable. We want our modules to be short so there is less scrolling. Please remove the `#---#`.

## Build system
CMake sounds like it would be a better build system (for a future pull request), but @newville and I don't see why this pull requires adding complexity to `setup.py`.

setup.py
carterbox: Oh! I see. You are relying on `shutil.copy2` to return three different error messages for each of the three errors you expect to occur. That's clever.

setup.py
jrmadsen: ## Ok, I hope this clears everything up

You cannot just add these lines to TomoPy. 

```python
    from build import buildtomopy
    os.chdir('config')
    build_libtomopy()
    os.chdir('..')
```

I know it seems like you can, you probably tested it and all the unit tests passed so everything seems fine. Now leave that directory and try to run something with tomopy. _**Everything will fail**_ because that "simple" addition does not result in libtomopy actually getting installed. THAT is why I need to do the copy. I did not just add in a bunch of "excess" code. It is entirely necessary to add those extra layers so that you know exactly where you need to copy libtomopy so that it gets installed. If you don't realize why this is the case it is because when separated into two different commands (build and install) libtomopy exists in `tomopy/sharedlibs` when install gets invoked so `libtomopy` is included when setuptools globs the files within `tomopy` -- which setuptools recognizes as the package folder since it matches the name specified in `setup(name="tomopy", ...)`. When everything is in one step, libtomopy does not exist (because it hasn't been built yet) so it is not included in the install manifest because it has absolutely zero way of know that file will exist eventually. Thus I had to add all this "Extension" code to get the path that I need in order to copy the compiled library into the folder that setuptools installs.

> Well, sort of. Setuptools seems to expect that an extensions includes python calls in C (or similar) code

setuptools makes no such assumption unless you are trying to use setuptools to compile code -- which it absolutely not the case here.

> It turns out that the compiled code in tomopy does not use Python.h at all and compiles straight C code to a shared library that could be used by (in principle) any language, and is used in tomopy by importing that shared library with ctypes. 

You only need `Python.h` if you are making Python interfaces to **_objects_**. Function calls are not objects and tomopy works the way it does without the need for `Python.h` because it only uses fundamental types -- `int`, `float`, etc. -- and numpy arrays which under the hood are just pointers to a starting memory address for a contiguous block of fundamental types (`int*`, `float*`, etc.).

> In that sense, libtomopy is as much a Python extension as libc is.

You can create Python bindings to functions defined in libc and that package would be a Python extension. It is not the library itself that is the extension, it is the combination of the bindings and the library.

> So, you can consider libtomopy an extension. OTOH, expecting that setuptools will automagically compile the code has proven pretty hard, especially on Windows. If you do want to alter the scripts to build libtomopy to build with CMake or to use setuptools.Extension, please make sure the resulting library compiles and works on Windows.

Again, none of my additions to setup.py have anything to do with setuptools compiling anything at all. Those PyCTest and TiMemory packages added to Travis are my packages. Both of these packages are:

- installable via `python setup.py install`
- use setuptools in general but use setuptools to compile absolutely nothing
- entirely use CMake to build the packages
- work with GCC, Clang, Intel, PGI, and MSVC compilers
    - MinGW is probably supported too, I just never bother because it is much better to use MSVC on Windows

In other words, I have done this sort of thing several times before.

jrmadsen:
@carterbox @newville Please see last post about why exactly these are necessary changes to setup.py. I will make them more "Pythonic" or whatever but your suggestions will break the `python setup.py install`

setup.py
carterbox: Thank you for that detailed explanation. It was helpful to understand your design decisions.

However, I believe we already solve the missing shared library problem by (1) using `pip` and (2) [including the shared libraries in `MANIFST.in`](https://blog.ionelmc.ro/2014/06/25/python-packaging-pitfalls/).

I support of my theory, I just now tested the use of `pip install .` to install `tomopy` from source into a clean `conda` environment after adding the following lines to `setup.py`:

```python
import os
from build import build_libtomopy

os.chdir('config')
build_libtomopy()
os.chdir('..')
```

Then, I checked what was installed in my environment's site-packages for `tomopy`, and I saw `libtomopy.dll` where I expected it to be. Finally, I uninstalled `tomopy` using `pip` and the folder I was viewing disappeared.

Is there a reason that you want/need to call `setup.py install` directly instead of using `pip`? [One of the reasons](https://stackoverflow.com/a/15731459) that using `pip` is better is that it makes uninstalling packages easier because it keeps track of which files were added. The alternative is deleting all of the files manually.

jrmadsen:
Yes. Anaconda and the ability for users to do standard things like call `python setup.py install` and have it work without having them  find out that installation only works through pip. I personally don't like or use pip bc in my opinion it is terrible for package management compared to Anaconda.

newville:
On Fri, Nov 2, 2018 at 9:52 PM Jonathan R. Madsen <notifications@github.com>
wrote:

> *@jrmadsen* commented on this pull request.
> ------------------------------
>
> In setup.py
> <https://github.com/tomopy/tomopy/pull/355#discussion_r230543088>:
>
> > +        src = TomoPyLibraryFile()
> +        src_base = os.path.basename(src)
> +        dst = join_path_list([extdir, "tomopy", "sharedlibs", src_base])
> +        print("")
> +        print('Library: {}'.format(src))
> +        print('Destination: {}'.format(dst))
> +        print("")
> +        try:
> +            import shutil
> +            shutil.copy2(src, dst)
> +        except Exception as e:
> +            print("Exception occurred copying '{}' to '{}'... {}".format(src, dst, e))
> +
> +
> +# ---------------------------------------------------------------------------- #
> +#
>
> Ok, I hope this clears everything up


Well, not completely.  First, I am not a tomopy devel, I'm just an
interested user who helped alter the build scripts to build tomopy on
Windows.  This is a recent change, but our group and downstream projects are
already relying on this feature.  My main concern and reason for
responding here
is that I would really like to see tomopy supported on Windows as well as
other platforms continue.

I don't really know what problem this PR is trying to solve.  It seems to
add dependencies in order to test with CMake.  I am not very familar with
CMake (built with it, never wrote my one build scripts with it), but have
no objections to it on technical merits.  Whether any build tool is used with
tomopy should be up to the main developers and maintainers. Then again, the
Makefiles used in tomopy are pretty darn simple.  Maybe the motivation for
this PR has been established and I missed the discussion.  That would be
OK, but it may be helpful to summarize that here so that the changes to the
code can be assessed relative to the actual goal.
FWIW, I may also not be reading this conversation on github, which can make
it challenging to follow the conversation well.

You cannot just add these lines to TomoPy.
>
>     from build import buildtomopy
>     os.chdir('config')
>     build_libtomopy()
>     os.chdir('..')
>
> I know it seems like you can, you probably tested it and all the unit
> tests passed so everything seems fine.
>

I struggle to imagine how having those lines at the top of `setup.py` could
be different from doing "python build.py && python setup.py build", which
(as near as I can tell) is what you are trying to replace.  If you are
worried about the success of the build and want an indication of that, you
could alter `build_tomopy.py` to return the exit status of running `make`,
and/or the full path of the created dynamic library.   That would be
working within the existing code rather than adding additional code to
`setup.py`.

Anyway, `python build.py && python setup.py install` should definitely
work.  If it does not, I would suggest you describe that as an Issue
*before* proposing a solution.


Now leave that directory and try to run something with tomopy. *Everything
> will fail* because that "simple" addition does not result in libtomopy
> actually getting installed. THAT is why I need to do the copy.
>

On the master branch, if you leave the top-level directory and expect to be
able to import tomopy, it will have have to tbe installed with "python
setup.py install" (which will install libtomopy), or placing that in a
custom Python path (which is sort of "custom install").  Are you expecting
different behavior?

I did not just add in a bunch of "excess" code. It is entirely necessary to
> add those extra layers so that you know exactly where you need to copy
> libtomopy so that it gets installed. If you don't realize why this is the
> case it is because when separated into two different commands (build and
> install) libtomopy exists in tomopy/sharedlibs when install gets invoked
> so libtomopy is included when setuptools globs the files within tomopy --
> which setuptools recognizes as the package folder since it matches the name
> specified in setup(name="tomopy", ...).
>

You certainly added more code than the one line "python build.py" to
`setup.py`.  It *seems* like the code you added is meant  to "build
libtomopy" -- is it meant to more than that?  If so, what is the goal?

When everything is in one step, libtomopy does not exist (because it hasn't
> been built yet) so it is not included in the install manifest because it
> has absolutely zero way of know that file will exist eventually.
>

Build libtomopy before running `setup()`.  It should not matter whether
this is done in the same Python process that runs `setup()` or in a
different process.  If you want to check that the build has succeeded, add
a check to `build.py`.


> Thus I had to add all this "Extension" code to get the path that I need in
> order to copy the compiled library into the folder that setuptools installs.
>

You do not need the Extension argument to `setup()`.  It might be possible
to use it (or some of the othe setuptools oddities like "eager_resources"),
but the current master branch does not use it and builds and installs
tomopy just fine without it.

Well, sort of. Setuptools seems to expect that an extensions includes
> python calls in C (or similar) code
>
> setuptools makes no such assumption unless you are trying to use
> setuptools to compile code -- which it absolutely not the case here.
>

Python Extensions are compiled code that are specially built to be
importable from Python.  https://docs.python.org/3/extending/building.html

It turns out that the compiled code in tomopy does not use Python.h at all
> and compiles straight C code to a shared library that could be used by (in
> principle) any language, and is used in tomopy by importing that shared
> library with ctypes.
>
> You only need Python.h if you are making Python interfaces to *objects*.
> Function calls are not objects and tomopy works the way it does without the
> need for Python.h because it only uses fundamental types -- int, float,
> etc. -- and numpy arrays which under the hood are just pointers to a
> starting memory address for a contiguous block of fundamental types (int*,
> float*, etc.)
>

Well, for sure the code for libtomopy doesn't use Python's API or create
Python objects within the C code.   It uses `ctypes`.  Ctypes is not
exactly limited to 'fundamental types', but that is not relevant here.

In that sense, libtomopy is as much a Python extension as libc is.
>
> You can create Python bindings to functions defined in libc and that
> package would be a Python extension. It is not the library itself that is
> the extension, it is the combination of the bindings and the library.
>

Python code that uses the ctypes module to load a shared library is pure
python code.  It is independent of Python version and system architecture
(at least in principle).  Tomopy builds libtomopy as a python-independent
shared library, installs that as an "extra package file", and then loads
that shared library at runtime with ctypes.  Tomopy is pure python, using a
pure C library.  I use this approach myself for a few different projects,
and I can strongly recommend it over writing Extension code.

So, you can consider libtomopy an extension. OTOH, expecting that
> setuptools will automagically compile the code has proven pretty hard,
> especially on Windows. If you do want to alter the scripts to build
> libtomopy to build with CMake or to use setuptools.Extension, please make
> sure the resulting library compiles and works on Windows.
>
> Again, none of my additions to setup.py have anything to do with
> setuptools compiling anything at all.
>

Yes, and this is why you do not *need* the Extension argument in the call
to `setup()`.  You do need the shared library to be built before `setup.py
install` is run so that the actual library can be installed.

Those PyCTest and TiMemory packages added to Travis are my packages. Both
> of these packages are:
>
>    - installable via python setup.py install
>    - use setuptools in general but use setuptools to compile absolutely
>    nothing
>    - entirely use CMake to build the packages
>    - work with GCC, Clang, Intel, PGI, and MSVC compilers
>       - MinGW is probably supported too, I just never bother because it
>       is much better to use MSVC on Windows
>
>
I have no objection to these goals or to CMake. But the last line raises a
concern for me.  I can tell you that I was not able to build libtomopy.dll
with MSVC on Windows due to unsupported C99 constructs.  It would probably
be possible to change the C code to comply with MSVC, but it seemed easier
and less disruptive to use MinGW (and it is easily added to the Anaconda
Python emvironment which is used and recommended for tomopy).  I can
certainly believe that someone more familar with MSVC and who think it is
much better to use MSVC than MinGW may be able to solve that issue with
C99.  I could not.

In other words, I have done this sort of thing several times before.
>

Great!  Again, my main concerns is that it work on Windows.  My secondary
concern is that it works the way the developers and maintainers want it to
work, as I want them to be able to focus on keeping tomopy awesome without
having to spend too much time tweaking build scripts.

carterbox:
@newville, Don't worry. We aren't going to abandon Windows support. 😄 The main purpose of this pull is to add benchmarking tests for high performance computing. This pull is unmerged because also includes a revision of the build system which is not related to the benchmarks.

@jrmadsen, The `MANIFST.in` solution should be working regardless of whether you choose to use `pip`. 😕 I think you've actually found a bug which probably does not requires wrapping the current build system with all that code.

We also like `conda`. `conda` is our preferred method for installing the pre-built `tomopy`. However, you'll notice that `meta.yml` uses `pip` to install from source because according to the python docs "[`pip` is the preferred installer program.](https://docs.python.org/3.6/installing/index.html)". `conda` cannot install packages from source.

jrmadsen:
As noted in the issue -- we were talking about two different situations. If you simply add:

```python
import subprocess as sp
sp.check_call((sys.executable, 'build.py'))
```
to the beginning of `setup.py` then `python setup.py install` installs correctly. This is perfectly inline with what I was saying earlier about `libtomopy.<ext>` existing when `setup(...)` is invoked. 
However, if you just do this:

```python
class TomoPyExtension(Extension):
    def __init__(self, name, sourcedir=''):
        Extension.__init__(self, name, sources=[])
        self.sourcedir = os.path.abspath(sourcedir)

class TomoPyBuild(build_ext, Command):

    def run(self):
        for ext in self.extensions:
            self.build_extension(ext)

    def build_extension(self, ext):
        sp.check_call((sys.executable, 'build.py'))

setup(
    # ... etc.
    ext_modules=[TomoPyExtension('libtomopy')],
    cmdclass=dict(build_ext=TomoPyBuild),
    # ... etc.
)
```
the installation does not include `libtomopy.<ext>` regardless of the `MANIFEST.in` setting. This is not a "bug" -- I explained why this happens previously.

Also, noted in the issue, this is part of a long-term strategy for extensions to TomoPy which will require
more complexity in the build system and the ability to enable/disable certain features that the latter provides via `setup.cfg`:

```python
[build_ext]
# customization settings
```

and their translation to command-line arguments to override the defaults

benchmarking/pyctest_tomopy_phantom.py
```diff
@@ -0,0 +1,282 @@
...
+import matplotlib
+import timemory
+import timemory.options as options
+import signal
+import numpy as np
+import time as t
+import pylab
+try:
+    from pyctest_tomopy_utils import *
+except:
```
carterbox: This should probably be `except ImportError:`.

benchmarking/pyctest_tomopy_phantom.py
```diff
@@ -0,0 +1,282 @@
...
+    if not os.path.exists(adir):
+        os.makedirs(adir)
+
+    if len(args.compare) == 0:
+        try:
+            import shutil
+            if os.path.exists(adir):
+                shutil.rmtree(adir)
+                os.makedirs(adir)
+        except:
```
carterbox: What exception are you trying to catch here?

benchmarking/pyctest_tomopy_phantom.py
```diff
@@ -0,0 +1,282 @@
...
+            pass
+    else:
+        try:
+            import shutil
+            if os.path.exists(adir):
+                shutil.rmtree(adir)
+                os.makedirs(adir)
+        except:
```
carterbox: What exception are you trying to catch here?

benchmarking/pyctest_tomopy_rec.py
```diff
@@ -0,0 +1,344 @@
...
+import numpy as np
+import collections
+
+import h5py
+import tomopy
+import dxchange
+import timemory
+try:
+    from pyctest_tomopy_utils import *
+except:
```
carterbox: This should probably be `except ImportError:`

jrmadsen:
They are not independent. The setup.py was modified to do one stage because I need to issue one command in pyctest_tomopy.py — where I set `pyctest.BUILD_COMMAND`. Otherwise I would have to call `python build.py` in the configuration stage and as a result, any the build warnings and/or errors would show up in “configure” section of CDash, not the “build” section.

carterbox:
OK. I see where `pytest` calls `python setup.py install`. I assume it records all the errors and warnings until that process returns, so if that call has a child process, then those will be recorded too.

Thus, if you add the following lines to `setup.py`:
```python
import subprocess as sp
sp.check_call((sys.executable, 'build.py'))

setup(...
```
Then the `build.py` is run a child of `setup.py`. Alternatively, you could `import build.py` and call `build_libtomopy()` directly.

jrmadsen:
@carterbox @dgursoy Also, whoever manages the Travis build system, please contact me so I can provide the CTEST_TOKEN so that the Travis builds can submit to CDash

carterbox:
@jrmadsen There are some bare exceptions in the pyctests. It's usually better to specify which exceptions you want to catch.

Also I'm opening another pull request to your branch with a temporary replacement for your custom Extension. I believe it fits all the necessary criteria for the present implementation of pyctests.

I'm also going push some commits which update our build infrastructure to remove the build step before the install step because those are now redundant.

jrmadsen:
Those exceptions are designed to catch non-critical errors that very likely won’t occur but if they do, they should simply be echoed, ignored, and execution should continue because they are not mission critical.

dgursoy:
Do we need to say anything about this on the docs?

carterbox:
No, the shared library is installed to the expected directory for all situations now.

dgursoy:
Thanks @jrmadsen for this PR, and @carterbox and @newville for the revision!

carterbox:
Actually, this breaks `python setup.py develop` because the two new installation directories are the same and Make doesn't like that. Let's fix that in a future build system.

jrmadsen:
I see what you are referring to but I don't get an error. No complaints and exit code is 0

jrmadsen:
Ok, fixed it. Simple:
```python
    if INSTALL_PREFIX is not None:
        cmd.append('install')
```
is now:

```python
    _PREFIX_PATH = os.path.abspath(INSTALL_PREFIX)
    _BINARY_PATH = os.path.abspath(os.path.join(os.getcwd(), ".."))
    if INSTALL_PREFIX is not None and _PREFIX_PATH != _BINARY_PATH:
        cmd.append('install')
```

PR #357, opened 2018-11-02
Misspelling in tomopy.rst

CJ-Wright:


carterbox:
Thanks for catching that, @CJ-Wright!

PR #366, opened 2018-11-09
BUG: Fix broken setup.py develop with build system update

carterbox:
The new build system (#355) broke setup.py develop. Fix provided by
@jrmadsen prevents Make from complaining when it is asked to
install the shared library to the same directory twice.

### DIFF
--- a/benchmarking/pyctest_tomopy_utils.py
+++ b/benchmarking/pyctest_tomopy_utils.py
@@ -19,7 +19,6 @@ import timemory.options as options
 import numpy as np
 import pylab
 import scipy.ndimage as ndimage
-import matplotlib.pyplot as plt
 import numpy.linalg as LA
 
 
--- a/build.py
+++ b/build.py
@@ -62,7 +62,10 @@ def build_libtomopy():
 
 def clean_libtomopy():
     """Clean libtomopy shared library for the current system."""
-    subprocess.check_call(('make', 'clean', '-f', get_makefile()))
+    if os.path.exists("Mk.config"):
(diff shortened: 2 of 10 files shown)


jrmadsen:
Also replace this:
```python
def clean_libtomopy():
    """Clean libtomopy shared library for the current system."""
    subprocess.check_call(('make', 'clean', '-f', get_makefile()))
```
with this:
```python
def clean_libtomopy():
    """Clean libtomopy shared library for the current system."""
    if os.path.exists("Mk.config"):
        subprocess.check_call(('make', 'clean', '-f', get_makefile()))
    else:
        print("Mk.config does not exist. Assuming nothing to clean...")
```
It wasn't causing an failures but PyCTest logs were polluted

carterbox:
@jrmadsen Do you want that "Mk.config does not exist. Assuming nothing to clean..." to go to `stderr` instead? `print` goes to `stdout`.

jrmadsen:
No its not an error really, PyCTest just calls clean by default whenever it starts up to make sure TomoPy is actually doing a fresh build and doesn't report zero build warnings, etc. because nothing actually got recompiled.

jrmadsen:
@dgursoy @carterbox can we get this PR closed? I've got some optimizations that depend on OpenMP SIMD instructions and thus require new build flags. The optimizations are producing around 1.2x to 1.25x speed up for all the iterative algorithms on all the architectures at NERSC (and I expect workstations will see the same degree of speed-up)

dgursoy:
Docs is broken after this merge.

dgursoy:
Docs is fine. It was a RTD issue.

PR #367, opened 2018-11-13
detector drift auto detection for aps_1id data 

KedoKudo:
add supporting functions (misc.npmath) and slits-based auto-alignment functions for tomography data from aps\_1id.


carterbox:
Hi @KedoKudo,

Thanks for proposing some new features for `tomopy`. It looks like your changes broke all of our tests. Check out the [tests section](https://tomopy.readthedocs.io/en/latest/devguide.html#running-the-tests) of our development documentation to read about how to run the tests on your local machine. Please give me a shout if you run into problems.

KedoKudo:
It seems to me that the tests failed because of the testing environment (the current testing environment cannot handle new __f-string formatting__,  and missing package __Pandas__).  On my local machine (python3), all tests passed without any issues (see [test.log](https://github.com/tomopy/tomopy/files/2581290/test.log) for more details).

prjemian:
That's right, the f formatting is a py2.6+ feature.  For robust code in py2.7 and py3.5+, refactor (as an example):

    f"report/{reportfn}.pdf"

to

    "report/{}.pdf".format(reportfn)

or (even more robust)

    os.path.join("report", "{}.pdf".format(reportfn))

prjemian:
The py36 build fails not for the *f-string* formatting but this error:

    Failure: ModuleNotFoundError (No module named 'pandas') ... ERROR

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,117 @@
...
+        Number of elements in the returning array
+
+    Returns
+    -------
+    pltX  : 1D np.array
+        plt data along x (data direction)
+    pltY  : 1D np.array
+        plt data long y (density direction)
+    """
+    if isinstance(data, pd.DataFrame):
```
carterbox: It looks like you only use `pandas` here on this one line. We try to minimize the number of dependencies that `tomopy` needs because it make installations smaller and may improve compatibility. Can you explain why your new feature requires the installation of `pandas`?

carterbox:
@KedoKudo,

Thanks in advance for you patience. As prjemian mentioned, `tomopy` targets python 2.7 and 3.5+ on Linux, macOS, and Windows. It may take a few days for your code to be fully reviewed. Expect that I will be asking a lot of questions and asking for more changes in order to ensure that the `tomopy` API doesn't break without good reason.

tomopy/misc/npmath.py
KedoKudo: Based on my experience, Pandas can be very handy when it comes to statistic analysis with a large dataset with many features/labels, which is something difficult to do with barebone numpy.  I would even recommend Tomopy to use pandas to handle tabulated data internally if possible.  However, if this is something that goes against Tomopy development policy, I can remove this function from the module.

prjemian: @KedoKudo : You are describing a feature enhancement for tomopy.  Sounds worthwhile.  Is it necessary for this PR?  Can the `pandas` requirement be separated so the merits can be discussed independent of this PR?

carterbox: It's not against development policy to introduce new packages, but in this particular case, it looks like the only reason you want to add `pandas` as a dependency is so you can check whether or you need to index data. I think you should refactor this way:
```python
# yours which requires pandas
calc_cummulative_dist(data, label='foo')

# refactored without pandas dependency
calc_cummulative_dist(data['foo'])
```

KedoKudo: @prjemian : That's a good point.  I decoupled the function dependencies on pandas.

tomopy/prep/alignment.py
```diff
@@ -443,3 +460,291 @@ def shift_images(prj, sx, sy):
...
+    #            )
+    
+    fig.savefig(os.path.join("report", "{}.pdf".format(reportfn)), 
+                transparent=True, 
+                bbox_inches='tight', 
+                pad_inches=0.1,
+                dpi=120,
+               )
+    
+    plt.close()  # close all to prevent memory leak
```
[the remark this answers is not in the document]
KedoKudo: report function is removed from Tomopy

tomopy/prep/alignment.py
```diff
@@ -49,17 +49,30 @@
...
 from tomopy.recon.algorithm import recon
-from tomopy.sim.project import project
+from tomopy.sim.project     import project
+from tomopy.misc.npmath     import gauss
+from tomopy.misc.npmath     import calc_affine_transform
+from tomopy.misc.npmath     import calc_cummulative_dist
+from scipy.signal           import medfilt2d
+from scipy.optimize         import curve_fit
+from scipy.ndimage          import affine_transform
+from scipy.ndimage          import shift
```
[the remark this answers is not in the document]
KedoKudo: import is condensed to a single line

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,111 @@
...
+    pad = lambda x: np.hstack([x, np.ones((x.shape[0], 1))])
+
+    # NOTE:
+    #   scipy affine_transform performs as np.dot(m, vec), 
+    #   therefore we need to transpose the matrix here 
+    #   to get the correct rotation
+    
+    # A, res, rank, s = np.linalg.lstsq(pad(pts_src), pad(pts_tgt))
+    # return A.T  
+    return np.linalg.lstsq(pad(pts_src), pad(pts_tgt), rcond=-1)[0].T
```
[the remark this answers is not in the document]
KedoKudo: alternative expression in comment removed

carterbox:
Thanks again, @KedoKudo for this pull request. It looks good to me, but I think @dgursoy had a question about something. He will ask or merge this pull request when he gets back from vacation.

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+def gauss(x, *p):
+    """
+    simple Guassian func used for quick curve fit
+
+    Parameters
+    ----------
+    x  :  np.1ndarray
+        1D array for curve fitting
+    p  :  parameter lis t
+        A,          mu,      sigma  =   p
```
dgursoy: unnecessary spacing after commas

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+        magnitude  center     std
+
+    Returns
+    -------
+    1d Gaussian distrobution evaluted at x with p
```
dgursoy: distrobution -> distribution

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    A, mu, sigma = p
+    return A*np.exp(-(x-mu)**2/(2.*sigma**2))
```
dgursoy: PEP8 fixes throughout the document

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+import numpy  as np
+import logging
+import warnings
+
+logger = logging.getLogger(__name__)
+
+
+__author__ = "Chen Zhang"
+__all__    = [ 'gauss',
+               'calc_cummulative_dist',
```
dgursoy: calc_ parts can be removed to shorten the function names.

scipy.stat uses discrete_cdf as a function name for cumulative distrubutions and we can maybe use the same.

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+]
+
+
+def gauss(x, *p):
+    """
+    simple Guassian func used for quick curve fit
+
+    Parameters
+    ----------
+    x  :  np.1ndarray
```
dgursoy: 1ndarray -> ndarray

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    -------
+    1d Gaussian distrobution evaluted at x with p
+    """
+    A, mu, sigma = p
+    return A*np.exp(-(x-mu)**2/(2.*sigma**2))
+
+
+def calc_cummulative_dist(data, steps=None):
+    """
+    Calculated the cumulative distribution from a data array without using
```
dgursoy: Can you add a sentence about what you refer to by saying binning?

Calculated -> Calculate
using can be dropped.
This is a more robust way of representing

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    binning.  
+    This provides a more robust way to representing the statistics of the 
+    incoming data without worrying about the binning/samping effect.
+
+    Parameters
+    ----------
+    data  :  1D np.array
```
dgursoy: 1D np.array -> np.ndarray

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+        1-D numpy array 
+    steps :  [ None | int ], optional
```
dgursoy: How the default value is calculated?

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+        Number of elements in the returning array
+
+    Returns
+    -------
+    pltX  : 1D np.array
```
dgursoy: pltX  : 1D np.array -> np.ndarray

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+        plt data along x (data direction)
+    pltY  : 1D np.array
+        plt data long y (density direction)
```
dgursoy: plt data long y -> Plotted data along y? or simply: Data along y (density) direction?

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    x = np.sort(data)
+
+    # check if list empty
```
dgursoy: Check if list is empty

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    if len(x) < 1:
```
dgursoy: len(x) == 0:

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+        return [], []
+
+    # subsamping if steps is speficiied and the number is smaller than the
```
dgursoy: speficiied -> specified

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    xx = np.tile(x, (2, 1)).flatten(order='F')
+    y = np.arange(len(x))
+    yy = np.vstack((y, y+1)).flatten(order='F')/float(y[-1])
+
+    return xx, yy
+
+
+def calc_affine_transform(pts_src, pts_tgt):
+    """
+    Use least square regression to calculate  the 2D affine transformation
```
dgursoy: redundant space after calculate

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+
+    # calculate the cumulative density
+    xx = np.tile(x, (2, 1)).flatten(order='F')
+    y = np.arange(len(x))
+    yy = np.vstack((y, y+1)).flatten(order='F')/float(y[-1])
+
+    return xx, yy
+
+
+def calc_affine_transform(pts_src, pts_tgt):
```
dgursoy: pts_src -> pts_source
pts_tgt -> pts_target

dgursoy: Why not use skimage.transform.AffineTransform instead?

tomopy/misc/npmath.py
```diff
@@ -0,0 +1,109 @@
...
+    """
+    # augment data with padding to include translation
+    pad = lambda x: np.hstack([x, np.ones((x.shape[0], 1))])
+
+    # NOTE:
+    #   scipy affine_transform performs as np.dot(m, vec), 
+    #   therefore we need to transpose the matrix here 
+    #   to get the correct rotation
+      
+    return np.linalg.lstsq(pad(pts_src), pad(pts_tgt), rcond=-1)[0].T
```
dgursoy: scipy.linalg can be faster.

tomopy/prep/alignment.py
```diff
@@ -49,17 +49,25 @@
 import numpy as np
 import logging
 import warnings
-from skimage import transform as tf
-from skimage.feature import register_translation
+import os
+
+from skimage                import transform              as tf
```
dgursoy: Could you remove the spaces before imports for a consistent style?

tomopy/prep/alignment.py
```diff
@@ -69,7 +77,11 @@
            'add_jitter',
            'add_noise',
            'blur_edges',
-           'shift_images']
+           'shift_images',
+           'find_slits_corners_aps_1id',
```
dgursoy: I'd try to find a way to drop aps_1id parts from these functions and provide a generic description to them. After-all these can be named as standard image processing functions. More comments/suggestions to follow when I check what they do.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+                               method='quadrant+', 
+                               medfilt2_kernel_size=3,
+                               medfilt_kernel_size=23,
+                               ):
+    """
+    Automatically locate the slit box location by its four corners.
+
+    Parameters
+    ----------
+    img                   :  np.2darray
```
dgursoy: Unnecessary spacing.

dgursoy: np.2darray -> np.ndarray

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        input 2D images
```
dgursoy: input can be dropped

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
     return prj
+
+
+def find_slits_corners_aps_1id(img, 
+                               method='quadrant+', 
+                               medfilt2_kernel_size=3,
+                               medfilt_kernel_size=23,
+                               ):
+    """
+    Automatically locate the slit box location by its four corners.
```
dgursoy: This needs more description here, because it seems this is very technique specific. Definition of the slits and its frame is needed. Also some explanation is needed about why you need corners of this box.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+
+    Parameters
+    ----------
+    img                   :  np.2darray
+        input 2D images
+    method                :  str,  ['simple', 'quadrant', 'quadrant+'], optional
```
dgursoy: How do simple, quadrant, quadrant+ differ from each other? Some explanation for the user is needed.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        method for auto detecting slit corners
+    medfilt2_kernel_size  :  int, optional 
+        2D median filter kernel size for noise reduction
+    medfilt_kernel_size   :  int, optional
```
dgursoy: I'd simply use kernel_size and explain in-line that it is for the median filter below to shorten the argument name.

p.s. This is also consistent with scipy.signal.medfilt

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+    Parameters
+    ----------
+    img                   :  np.2darray
+        input 2D images
+    method                :  str,  ['simple', 'quadrant', 'quadrant+'], optional
+        method for auto detecting slit corners
+    medfilt2_kernel_size  :  int, optional 
+        2D median filter kernel size for noise reduction
+    medfilt_kernel_size   :  int, optional 
+        1D median filter kernel size for noise reduction
```
dgursoy: What is the function of 1D and 2D median filters?

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+    
+    Returns
+    -------
+    tuple
+        autodetected slit corners (counter-clockwise order)
```
dgursoy: In 2D or in 3D?

tomopy/prep/alignment.py
```diff
@@ -49,17 +49,25 @@
...
-from skimage.feature import register_translation
+import os
+
+from skimage                import transform              as tf
+from skimage.feature        import register_translation
 from tomopy.recon.algorithm import recon
-from tomopy.sim.project import project
+from tomopy.sim.project     import project
+from tomopy.misc.npmath     import gauss, calc_affine_transform, calc_cummulative_dist
+from scipy.signal           import medfilt, medfilt2d
```
dgursoy: Isn't medfilt2d a subset of medfilt? So, why not just use medfilt?

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+    medfilt_kernel_size   :  int, optional 
+        1D median filter kernel size for noise reduction
+    
+    Returns
+    -------
+    tuple
+        autodetected slit corners (counter-clockwise order)
+        (upperLeft, lowerLeft, lowerRight, upperRight)
+    """
+
```
dgursoy: Redundant blank

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+                    kernel_size=medfilt2_kernel_size,
+                    )
+    rows, cols = img.shape
+    
+    # simple method is simple, therefore it stands out
+    if method.lower() == 'simple':        
+        # assuming a rectangle type slit box 
+        col_std = medfilt(np.std(img, axis=0), kernel_size=medfilt_kernel_size)
+        row_std = medfilt(np.std(img, axis=1), kernel_size=medfilt_kernel_size)
+        # NOTE: in the tiff img
```
dgursoy: Some of the notes can be shifted to or duplicated in the function docstring.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        # predefine all quadrants
+        # Here let's assume that the four corners of the slit box are in the 
+        # four quadrant defined by the center of the image
+        # i.e.
+        #  uppper left quadrant: img[0     :cnt[1], 0     :cnt[0]]  => quadarnt origin =  (0,           0)
+        #  lower  left quadrant: img[cnt[1]:      , 0     :cnt[0]]  => quadarnt origin =  (cnt[0],      0)
+        #  lower right quadrant: img[cnt[1]:      , cnt[0]:      ]  => quadarnt origin =  (cnt[0], cnt[1])
+        #  upper right quadrant: img[0     :cnt[1], cnt[0]:      ]  => quadarnt origin =  (0,      cnt[1])
+        cnt = [int(cols/2), int(rows/2)]  # center of image that defines FOUR quadrants
+        Quadrant  = namedtuple('Quadrant', 'img col_func, row_func')
```
dgursoy: vars should be all lowercase

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        quadrants = [Quadrant(img=img[0:cnt[1], 0:cnt[0]], col_func=np.argmax, row_func=np.argmax),  # upper left,  1st quadrant
+                     Quadrant(img=img[cnt[1]: , 0:cnt[0]], col_func=np.argmax, row_func=np.argmin),  # lower left,  2nd quadrant
+                     Quadrant(img=img[cnt[1]: , cnt[0]: ], col_func=np.argmin, row_func=np.argmin),  # lower right, 3rd quadrant
+                     Quadrant(img=img[0:cnt[0], cnt[1]: ], col_func=np.argmin, row_func=np.argmax),  # upper right, 4th quadrant
+                    ]
+        # the origin in each quadrants ==> easier to set it here
+        quadrantOrigins = np.array([[0,           0],  # upper left,  1st quadrant
```
dgursoy: vars should be all lowercase

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+                                    [0,      cnt[1]],  # lower left,  2nd quadrant
+                                    [cnt[0], cnt[1]],  # lower right, 3rd quadrant
+                                    [cnt[1],      0],  # upper right, 4th quadrant
+                                   ])
+        # init four corners
+        cnrs = np.zeros((4, 2))
+        if method.lower() == 'quadrant':
+            # the standard quadrant method
+            for i, q in enumerate(quadrants):
+                cnrs[i,:] = np.array([q.col_func(np.gradient(medfilt(np.std(q.img, axis=0), kernel_size=medfilt_kernel_size))),  # x is col_idx
```
dgursoy: Long line?

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+    """
+
+
+    img = medfilt2d(np.log(img.astype(np.float64)),
+                    kernel_size=medfilt2_kernel_size,
+                    )
+    rows, cols = img.shape
+    
+    # simple method is simple, therefore it stands out
+    if method.lower() == 'simple':
```
dgursoy: I'd refactor this if else conditions by creating and calling simple, quadrant and quadrant+ functions here.

This would also give a chance to explain the purpose of each in their own function docs.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        else:
+            raise NotImplementedError("Available methods are: simple, quadrant, quadrant+")
+    
+    # return the slit corner detected
+    return cnrs
+
+
+def calc_slit_box_aps_1id(slit_box_corners, inclip=(1,10,1,10)):
+    """
+    Calculate the clip box based on given slip corners
```
dgursoy: dot at the end of sentence

dgursoy: What is a clip box? What is a slit? Some definition of these is appreciated because it seems domain specific.

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+
+    Parameters
+    ----------
+    slit_box_corners  :  np.2darray
```
dgursoy: slit_box_corners  :  np.2darray -> slit_box_corners : np.2darray

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+        four corners of the slit box 
+    inclip            :  tuple, optional
+        Extra inclipping to avoid clipping artifacts
+
+    Returns 
+    -------
+    np.3darray
+        tomopy images stacks without regions outside slits
+    """
+    xl,xu,yl,yu = calc_slit_box_aps_1id(slit_box_corners, inclip=inclip)
```
dgursoy: spaces after commas

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+            # NOTE:
+            #  This is particular slow, need optimization
+            imgstacks[n_img,:,:] = affine_transform(imgstacks[n_img,:,:],                # input image
+                                                    img_correct_F[n_img,0:2,0:2],        # rotation matrix
+                                                    offset=img_correct_F[n_img,0:2,  2], # offset vector
+                                                   )
+    # convert proj_cnrs to np.array
+    proj_cnrs = np.stack(proj_cnrs, axis=0)
+
+    return imgstacks, proj_cnrs, img_correct_F
```
dgursoy: vars all lowercase

tomopy/prep/alignment.py
```diff
@@ -443,3 +455,240 @@ def shift_images(prj, sx, sy):
...
+                                              )
+        proj_cnrs.append(proj_cnr)
+
+        # calcualte the affine transfomration required
+        img_correct_F[n_img,:,:] = calc_affine_transform(proj_cnr, 
+                                                         slit_cnr_ref,
+                                                        )
+
+        # adjust the image
+        if np.linalg.norm(img_correct_F[n_img,0:2,0:2] - np.eye(2)) < 1e-4:
```
dgursoy: PEP8 fixes

tomopy/misc/npmath.py
KedoKudo: The default (steps=None) will use the whole data.  It is close to considering using bin_size=1 or bins=len(data).

tomopy/misc/npmath.py
KedoKudo: skimage.transform.AffineTransform() is used to transform an image with a given 2D affine transformation matrix.  This function generates the affine transformation matrix with given sets of points.

tomopy/prep/alignment.py
KedoKudo: The four slits that form a binding box is the current setup at aps_1id, which reduce the illuminated region on the detector.   Since the slits are stationary, they can serve as a reference to check detector drifting during the scan.  Technically, the four slits should be used to find the transformation matrix (not necessarily affine) to correct the image.  However, since we are dealing with 2D images with very little distortion, I use affine transformation matrices for approximation.  Therefore the "four corners" are used instead of all four slits.

tomopy/prep/alignment.py
KedoKudo: I think the prefix here is necessary as they are intended for different median filters for different purposes.  I am currently working on an adaptive method, which might be able to get away using a universal kernel_size.  However, that is for a different pull request, and I do have a few bugs to figure out before I can make the PR.

tomopy/prep/alignment.py
KedoKudo: 2D median filter:  remove noise points in the image due to defect defects

1D median filter:  reduce the local fluctuation of the 1D gradient profile to avoid miss-identification of the slit location

tomopy/prep/alignment.py
KedoKudo: 2D, four corners stacked in counter-clockwise order, starting from the upper left corner (following the raster scan coordinate system convention).

tomopy/prep/alignment.py
KedoKudo: medfilt2d is more suitable for processing image

tomopy/prep/alignment.py
KedoKudo: I use the namedtuple() here as a shorthand for an in-line class.

tomopy/prep/alignment.py
KedoKudo: autopep8 fix this problem by breaking it into multiple lines.

tomopy/prep/alignment.py
KedoKudo: The functionality of each method is added to the docstring.  

As for refactoring, I have to argue that breaking them into separate functions would make the code less readable and difficult to maintain.

tomopy/prep/alignment.py
KedoKudo: ASCII art is already added in the docstring.

tomopy/prep/alignment.py
KedoKudo: img_correct_F tells people that this is a matrix (M usually represents a matrix/tensor, whereas m represents either a vector or a scalar).  I believe this is better than writing it as img_correct_f or img_correct_matrices.

tomopy/misc/npmath.py
dgursoy: Ah, ok. Could you add this to the docstring?

tomopy/misc/npmath.py
dgursoy: Then, I'd rename the function something like affine_transform_matrix. Otherwise it is counter-intuitive.

tomopy/prep/alignment.py
dgursoy: If this is the case, I think it is fine. I was trying to learn of anyone else would need these functions. An alternative would be to use the technique as a descriptor rather than the beamline, because beamlines can change in time.

tomopy/misc/npmath.py
prjemian: For the purposes of automatically documenting code (such as [autosummary](http://www.sphinx-doc.org/en/master/usage/extensions/autosummary.html)), please reconsider the content format _in_ the doc strings.  The first line should be a summary *all on one line* with vertical white space for the remaining content.

In this case, change to this

```
+    """
    Calculate the cumulative distribution from a data array without using binning.  

    This provides a more robust way to representing the statistics of the 
    incoming data without worrying about the binning/sampling effect.

    Parameters
...
```

tomopy/prep/alignment.py
dgursoy: This is a nice explanation that we can use in the doc. :-)

tomopy/prep/alignment.py
dgursoy: I think it is fine then.

tomopy/prep/alignment.py
dgursoy: I think now I agree with you. I assume no other function will call these functions, so they better stay where they are right now.

tomopy/misc/npmath.py
KedoKudo: I updated the docstring accordingly

tomopy/misc/npmath.py
KedoKudo: The prefix calc_  already indicates this is a calculation, not an execution.  Also using calc_ instead of get_ implies that this is not an o(1) calculation.  If I were to implement the execution part myself, I would probably name it "do_affine_transform" instead "affineTransform".

tomopy/prep/alignment.py
KedoKudo: If the current setup here at 1-ID had a dedicated name, I would most likely use that instead of the beamline ID.  Unfortunately, I am not aware of whether this kind of setup has its own name, which leaves us with the beamline ID as the only option for now.

tomopy/prep/alignment.py
KedoKudo: I add the comment into the docstring as a secondary description.

KedoKudo:
@dgursoy I have resolved the conversations that do not require further feedback.  A few extra docstring edits have been made to address the new comments posted today.

dgursoy:
@KedoKudo Thanks! All looks good.

PR #368, opened 2018-11-28
Warnings fixes + optimizations + floating-point consistency + clang-format

jrmadsen:
- Warning fixes
    - Fixed sign comparison warnings in src/morph.c
- Optimizations in src/utils.c via SIMD instructions
    - enable with `export CFLAGS="-fopenmp"` before build
        - this compiler flags will eventually be tested for availability and automatically applied when CMake build system is implemented
    - recorded speed-up on Cori-KNL, Cori-Haswell, Edison: ~1.25x
    - Optimized functions
        - calc_quadrant
        - calc_coords
        - trim_coords
        - sort_intersections
        - calc_dist
- floating-point consistency
    - replaced div by 2 with multiply by 0.5f (faster)
    - replaced double constants with float constants
    - replaced fmod with fmodf
- clang-format
    - created .clang-format file (requires clang-format v6.0+)
        - run: `clang-format -i $(find . -type f | egrep '\.h$|\.c$')` to apply formatting in-place

### DIFF
--- a/src/utils.c
+++ b/src/utils.c
@@ -75,7 +75,7 @@ preprocessing(int ry, int rz, int num_pixels, float center, float* mov,
         gridy[i] = -rz * 0.5f + i;
     }
 
-    *mov = ((float) num_pixels - 1) * 0.50f - center;
+    *mov = ((float) num_pixels - 1) * 0.5f - center;
     if(*mov - floor(*mov) < 0.01f)
     {
         *mov += 0.01f;
--- a/src/utils.c
+++ b/src/utils.c
@@ -1,342 +1,459 @@
 // Copyright (c) 2015, UChicago Argonne, LLC. All rights reserved.
 
-// Copyright 2015. UChicago Argonne, LLC. This software was produced 
-// under U.S. Government contract DE-AC02-06CH11357 for Argonne National 
-// Laboratory (ANL), which is operated by UChicago Argonne, LLC for the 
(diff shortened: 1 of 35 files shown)


dgursoy:
Conflict due to #366 merge.

src/utils.c
```diff
@@ -1,342 +1,459 @@
...
+    for(int i = 0; i <= rz; ++i)
     {
-        gridy[i] = -rz/2.+i;
+        gridy[i] = -rz * 0.5f + i;
     }
 
-    *mov = ((float)num_pixels-1)/2.0-center;
-    if(*mov-floor(*mov) < 0.01) {
-        *mov += 0.01;
+    *mov = ((float) num_pixels - 1) * 0.50f - center;
```
carterbox: There's a trailing zero here that's not necessary.

jrmadsen: Fair point about separate commit for formatting. Normally I would have but wasn't paying attention to that when I rebased bc I was more concerned about ensuring any the MPI stuff wasn't included. The only C code of import affected by the optimizations was the functions listed in utils.c. For the algorithms it was FP consistency

jrmadsen:
If you want to deny this PR, I can re-do the rebase so the clang-format is separate.

carterbox:
That's not necessary. I already read through it all. I'm only requesting removal of that trailing zero.

PR #369, opened 2018-11-29
Fix broken conda build

carterbox:
Updates the install directions documentation.
Fixes unsatisfiable dependencies problem when building with conda.
h5py and pywavelets are no longer needed?


doc/source/install.rst
```diff
@@ -7,13 +7,25 @@ This section covers the basics of how to download and install TomoPy.
...
+-------------
+
+No issues with the current build system have been reported.
+
+Importing TomoPy
+----------------
+
+When importing, it is best to import TomoPy before importing numpy.
+See `this thread <https://github.com/tomopy/tomopy/issues/178>`_ for details.
+
```
dgursoy: Because we stopped FFTW as a dependency, Importing TomoPy section can be removed. I'd put this part under Common issues as a remark for users, who use FFTW instead of MKL-FFT.

dgursoy:
Thanks @carterbox. Before we merge, could you also set the versions in ```VERSION``` and ```tomopy/__init__.py``` files to 1.3.0?

Issue #371, opened 2018-12-04
windows build broken due to missing `install`


newville:
Trying to build from current master with Anaconda Python 3.6 on Windows (64 bit), gives

C:\Users\xas_user\Documents\GitHub\tomopy>python setup.py build
running build
running build_py
running egg_info
writing tomopy.egg-info\PKG-INFO
writing dependency_links to tomopy.egg-info\dependency_links.txt
writing top-level names to tomopy.egg-info\top_level.txt
reading manifest file 'tomopy.egg-info\SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no previously-included files matching '__pycache__' found under directory '*'
warning: no previously-included files matching '*.pyc' found under directory '*'
warning: no previously-included files matching '*.o' found under directory 'config'
writing manifest file 'tomopy.egg-info\SOURCES.txt'
copying tomopy\sharedlibs\libtomopy.dll -> build\lib.win-amd64-3.6\tomopy\sharedlibs
running build_ext
INFO:__main__:Config for Windows
INFO:__main__:COMPILER_DIR is C:\Users\xas_user\AppData\Local\Continuum\Anaconda3\MinGW\bin\gcc.exe
INFO:__main__:Config output:
## generated by build.py  Tue Dec  4 09:24:58 2018
COMPILER_DIR   = C:/Users/xas_user/AppData/Local/Continuum/Anaconda3/MinGW/bin
SHAREDLIB      = libtomopy.dll
ARCH_TARGET    =
LINK_LIB       = -LC:/Users/xas_user/AppData/Local/Continuum/Anaconda3 -LC:/Users/xas_user/AppData/Local/Continuum/Anaconda3/Libra
ry/bin -LC:/Users/xas_user/AppData/Local -LC:/Windows/System32
INCLUDE        = -IC:/Users/xas_user/Documents/GitHub/tomopy/include -IC:/Users/xas_user/AppData/Local/Continuum/Anaconda3/include
 -IC:/Users/xas_user/AppData/Local/Continuum/Anaconda3/Library/include
CONDA_COMPAT   =
INSTALL_PREFIX = C:/Users/xas_user/Documents/GitHub/tomopy/build/lib.win-amd64-3.6
####

C:/Users/xas_user/AppData/Local/Continuum/Anaconda3/MinGW/bin/gcc.exe -mdll -O -DMS_WIN64 -DUSE_MKL -Wall -IC:/Users/xas_user/Docu
[...]

PR #374, opened 2018-12-06
BLD: Use python functions to replace `make install` and `make clean` in build.py

newville:
This addresses #371  by using `shutil.copy()`, `os.chmod()`, and `os.unlink()` instead of relying on `make install` to run an external `install` program (and `make clean` to run an external `rm` program).   These external programs may not be available on all platforms and environments.

### DIFF
--- a/build.py
+++ b/build.py
@@ -12,6 +12,8 @@ from os.path import join as pjoin
 import logging
 import sys
 import subprocess
+import glob
+import shutil
 import time
 import logging
 import argparse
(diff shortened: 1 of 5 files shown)


carterbox:
I'm won't merge this pull because when running
```bash
python setup.py build
```

`libtomopy.so` is not copied to `build/lib.linux-x86_64-3.6/tomopy/sharedlibs/`

Thus, I believe that the shared library would not be installed to the final destination, and this patch does not correctly replace the install command in the make file.

newville:
@carterbox Oh, sorry about that.  Fixed now.

carterbox:
@newville I'm merging this PR because I think there may be other non-build related changes before this build system is replaced with CMake + scikit-build. Effectively, this is a temporary patch until then.

PR #375, opened 2018-12-07
WIP: CMake build system for TomoPy

jrmadsen:
## THIS PULL REQUEST IS A WORK IN PROGRESS!

- Uses `scikit-build` to run CMake
- Adds direct configuration options, e.g. `python setup.py --enable-arch --enable-gperf build`
- Has not been tests on Windows yet
- Has not deleted old build system yet


jrmadsen:
@carterbox I opened this PR early to start the feedback and testing

jrmadsen:
Python 3.5 + scikit-build has been failing for me for a while and it is an issue with scikit-build itself. I expect I will file an issue with them.

Also, there are some things that are partially integrated. For example:

- `TOMOPY_USE_TBB` option in CMake ultimately does nothing but eventually, I will create some C that, when enabled, will implement the TBB scalable replacements to `malloc`, `calloc`, etc. 
- `TOMOPY_USE_GPERF` option in CMake will enable uploading profiles within PyCTest
- `TOMOPY_USE_ITTNOTIFY` option in CMake will enable pausing/resuming VTune profiler

jrmadsen:
@newville This should solve the build system issues. Once integrated, there will be no need for `install.exe`, etc. and, actually, it may even get TomoPy working with the native Visual Studio compilers. This is not currently the case because the VS compilers don't support `float _Complex` in C and instead, Windows has their own `_Fcomplex` struct. This struct has a different assignment and initialization scheme so gridrec fails to compile. I have been told, however, there is a compiler flag to request POSIX compliance and am looking into it.

jrmadsen:
I have submitted a PR to `scikit-build` and spoken with the developers. The issue with Python 3.5 (and also Python 3.7) should be fixed soon.

newville:
@jrmadsen

> @newville  This should solve the build system issues. Once integrated, there will be no need for install.exe, etc. and, actually, it may even get TomoPy working with the native Visual Studio compilers. This is not currently the case because the VS compilers don't support float _Complex in C and instead, Windows has their own _Fcomplex struct. This struct has a different assignment and initialization scheme so gridrec fails to compile. I have been told, however, there is a compiler flag to request POSIX compliance and am looking into it.

Oh, my understanding is that MSVC does not support `float _Complex`.  If there were a compiler flag for that, I think it could be used with the current build scripts too.  Altering gridrec would be another option, but in September using MinGW on Windows seemed less disruptive (and Windows only).  I have not had or heard of problems with the MinGW compiled code, but given recent discussions, I should be clear that I do not claim that 'works for me' demonstrates that there are no problems.  Have you had issues with tomopy on Windows?

I have very little experience with CMake.  I am sure there are good reasons to use it, but I cannot help but notice that this PR would add ~2500 lines of CMake build files (and a dependency on scikit-build). For a project of this size, that seems like a lot of added code, and in a new language.  Can that be streamlined or simplified in any way?

PR #374 removes the use of external `install` and `rm` programs on all platforms by using functions from python standard library.  The build scripts there total ~200 lines of Python and ~65 lines of Makefile fragment.  That seems to be 10x smaller than the CMake scripts in this PR.  As an interested user, I hope the tomopy developers are able to focus most of their attention on the algorithms in the code, and not on maintaining the build system.

Hope that helps...

carterbox:
@jrmadsen Thanks so much! I was wondering if you had started this. One of the custom modules you added is for finding FFTW. Do you still use FFTW? @dgursoy I though we had switched to MKL.

>  I am sure there are good reasons to use it,...

@newville The number of new lines in this build system upgrade does seem like a lot, but most of it is custom CMake modules to determine automatically the appropriate build flags for lots of different compilers and to find libraries on the host system. The switch CMake is really necessary because we want to target multiple platforms and architectures. Specifically, we used to only target Unix workstations, and now we are including Windows workstations and a variety of Unix supercomputers. Most of our users will not notice this transition because they do not compile from source, so you're unique. Thank you for your continued feedback. I'm sorry in advance for disrupting your builds, but we will continue to make a good faith effort to test build changes before merging them to master (that's why @jrmadsen opened this pull early!). I will also try to keep the installation docs updated.

>I hope the tomopy developers are able to focus most of their attention on the algorithms in the code

@newville Some developers of TomoPy are interested in high-performance-computing applications, so that is their main focus. If there are some algorithmic deficiencies that you would like addressed. Please open an Issue.

newville:
@carterbox   OK.  That at least makes it seem like CMake is something you want to use and are comfortable with.  Perhaps you'll forgive me for suspecting that support for Windows would have not ever have happened if a CMake build system had been in place a few months back.

jrmadsen:
@carterbox There is still FFTW code in gridrec so I included the FindFFTW.cmake as a fallback if MKL cannot be found. In `Options.cmake` you will see the `TOMOPY_USE_MKL` is `ON` by default and in `Packages.cmake`, you will see FFTW is only used if MKL was explicitly disabled or not found:

```
if(NOT TOMOPY_USE_MKL OR NOT MKL_FOUND)
    find_package(FFTW3 COMPONENTS single threads_single)
    ...
```

jrmadsen:
@carterbox @dgursoy Could one of y'all re-trigger the Travis build? I checked and the scikit-build in conda-forge has been updated so it should be working now

jrmadsen:
@newville CMake is designed for cross-platform support (the name is "short" for `Cross-platform Make`). If anything, it should make supporting Windows easier. For example, in the build system, the copy to `tomopy/sharedlibs/` from the build directory is just:

```
add_custom_command(TARGET libtomopy POST_BUILD
    COMMAND ${CMAKE_COMMAND} -E copy $<TARGET_FILE:libtomopy>
        ${PROJECT_SOURCE_DIR}/tomopy/sharedlibs/)
```

This just mean after `libtomopy` is built, expand `$<TARGET_FILE:libtomopy>` to the build-tree output path to `libtomopy.dll` on Windows, path to `libtomopy.dylib` on macOS, etc. and CMake handles the platform specifics of how to do the copy (i.e. your PR #374).

jrmadsen:
@dgursoy @carterbox Should we remove the FFTW code from gridrec and CMake support for finding it?

I think it is broken: `python setup.py install -- -DTOMOPY_USE_MKL=OFF && nosetests` results in:

```
test_circ_mask (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_gaussian_filter (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_median_filter (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_remove_nan (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_remove_neg (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_remove_outlier (test.test_misc.test_corr.ImageFilterTestCase) ... ok
test_downsample (test.test_misc.test_morph.MorphingTestCase) ... ok
test_sino_360_to_180 (test.test_misc.test_morph.MorphingTestCase) ... ok
test_upsample (test.test_misc.test_morph.MorphingTestCase) ... ok
test_baboon (test.test_misc.test_phantom.TestDistributeJobs) ... /opt/conda/3/envs/tomopy/lib/python3.6/site-packages/skimage/transform/_warps.py:110: UserWarning: Anti-aliasing will be enabled by default in skimage 0.15 to avoid aliasing artifacts when down-sampling images.
  warn("Anti-aliasing will be enabled by default in skimage 0.15 to "
ok
test_barbara (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_cameraman (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_checkerboard (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_lena (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_peppers (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_shepp2d (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_shepp3d (test.test_misc.test_phantom.TestDistributeJobs) ... ok
test_normalize (test.test_prep.test_normalize.NormalizationTestCase) ... ok
test_normalize_bg (test.test_prep.test_normalize.NormalizationTestCase) ... ok
test_normalize_nf (test.test_prep.test_normalize.NormalizationTestCase) ... ok
test_retrieve_phase (test.test_prep.test_phase.TestDistributeJobs) ... ok
test_remove_stripe_fw (test.test_prep.test_stripe.StripeRemovalTestCase) ... ok
test_remove_stripe_ti (test.test_prep.test_stripe.StripeRemovalTestCase) ... ok
test_art (test.test_recon.test_algorithm.ReconstructionAlgorithmTestCase) ... ok
test_bart (test.test_recon.test_algorithm.ReconstructionAlgorithmTestCase) ... ok
test_fbp (test.test_recon.test_algorithm.ReconstructionAlgorithmTestCase) ... ok
test_grad (test.test_recon.test_algorithm.ReconstructionAlgorithmTestCase) ... ok
test_gridrec (test.test_recon.test_algorithm.ReconstructionAlgorithmTestCase) ... Segmentation fault: 11
```

newville:
@jrmadsen  
> If anything, it should make supporting Windows easier.

OK, if CMake is what everyone else wants, I am not opposed but probably won't be able to help much.

dgursoy:
> Do you still use FFTW? @dgursoy I though we had switched to MKL.

> @dgursoy @carterbox Should we remove the FFTW code from gridrec and CMake support for finding it?

We can drop the FFTW (and PyFFTW) from the dependency list (as part of #377). TomoPy is shipped with MKL by default and if a user wants to use FFTW or numpy.fft they are free to do, but they need to have them installed. 

Also see this module for calling FFT operations within TomoPy Python modules:
https://github.com/tomopy/tomopy/blob/master/tomopy/util/misc.py

carterbox:
@dgursoy We're aware that FFTW is a choice for FFT library, and it's already not an install requirement. What we're asking is whether we can remove it as a choice so that the only *two* options for FFT are MKL and numpy.

MKL is easily available for download through conda, and we don't even test to make sure that FFTW is still working for us. Only the MKL implementation is tested in our CI tests. This is probably why as @jrmadsen pointed out, gridrec  fails when he tries to use FFTW instead of MKL. Keeping FFTW around requires us to maintain more code and fix the apparently broken FFTW implementation (which no one even noticed because most of our users install using conda which uses MKL).

If we continue to support FFTW, we definitely should be testing it, and we should probably be offering a conda build variant in which PyFFTW is required *instead* of MKL. (e.g. NumPy offers MKL and OpenBLAS variants).

dgursoy:
@carterbox We can drop supporting FFTW.

jrmadsen:
@carterbox Here is a thought w.r.t. to how we have two copies of the compiled library (one in source tree and one in build tree). 

If we replace `tomopy/__init__.py`:

```python
from __future__ import (absolute_import, division, print_function,
                        unicode_literals)

__version__ = '1.3.0'

from ._fft_loader import fft_impl
from tomopy.allocator import *
from tomopy.misc.corr import *
from tomopy.misc.morph import *
from tomopy.misc.phantom import *
from tomopy.prep.alignment import *
from tomopy.prep.normalize import *
from tomopy.prep.phase import *
from tomopy.prep.stripe import *
from tomopy.recon.wrappers import *
from tomopy.recon.algorithm import *
from tomopy.recon.rotation import *
from tomopy.recon.vector import *
from tomopy.recon.acceleration import *
from tomopy.sim.project import *
from tomopy.sim.propagate import *
from tomopy.util.mproc import set_debug

import logging
logging.getLogger(__name__).addHandler(logging.NullHandler())
```

with `tomopy/__init__.py.in` (CMake template file -- CMake will replace `@PROJECT_VERSION@` with the contents of the `VERSION` file):

```python
from __future__ import (absolute_import, division, print_function,
                        unicode_literals)

__version__ = '@PROJECT_VERSION@'

from ._fft_loader import fft_impl
from tomopy.allocator import *
from tomopy.misc.corr import *
from tomopy.misc.morph import *
from tomopy.misc.phantom import *
from tomopy.prep.alignment import *
from tomopy.prep.normalize import *
from tomopy.prep.phase import *
from tomopy.prep.stripe import *
from tomopy.recon.wrappers import *
from tomopy.recon.algorithm import *
from tomopy.recon.rotation import *
from tomopy.recon.vector import *
from tomopy.recon.acceleration import *
from tomopy.sim.project import *
from tomopy.sim.propagate import *
from tomopy.util.mproc import set_debug

import logging
logging.getLogger(__name__).addHandler(logging.NullHandler())
```

and run
```cmake
configure_file(${PROJECT_SOURCE_DIR}/tomopy/__init__.py.in 
    ${PROJECT_BINARY_DIR}/tomopy/__init__.py 
    @ONLY)
``` 
within CMake, then we no longer would have to output the library to `tomopy/sharedlibs/libtomopy.{so,dylib,dll}` in the source tree because Python would no longer recognize the `tomopy/` directory in the source tree as a package and `import tomopy` would pick up the tomopy installed via `pip install .` or `python setup.py install`.

carterbox:
@jrmadsen, This is what I am understanding: The problem you are trying to solve is that `libtomopy` is copied to too many places. The only place it needs to be is `${INSTALL_DIR}/tomopy/sharedlibs`. `${INSTALL_DIR}` could be:
  - the build directory if we run `python setup.py build`
  - somewhere on the Python path if we run `python setup.py install`
  - the source directory if we run `python setup.py develop`.

The extra place that it does not need to be is `${PROJECT_SOURCE_DIR}/tomopy/sharedlibs`.

The reason that (in the past) we put `libtomopy` in `${PROJECT_SOURCE_DIR}/tomopy/sharedlibs` was that we wanted `setuptools` to copy `libtomopy` to `${INSTALL_DIR}/tomopy/sharedlibs`. However, since we are now using CMake, we can just have CMake move `libtomopy` directly to `${INSTALL_DIR}/tomopy/sharedlibs` because it knows the value of  `${INSTALL_DIR}`.

Thus, I am confused. Why do we need to trick Python into thinking that the `${PROJECT_SOURCE_DIR}/tomopy/` directory is not a package? If we want to use the package installed by `python setup.py`, then we only need to change the working directory (e.g. run tests from inside `${PROJECT_SOURCE_DIR}/tests/` instead of inside `${PROJECT_SOURCE_DIR}/`)?

Also, I do really like the idea of using CMake to reduce the number of places that the version number needs to be updated.

carterbox:
@jrmadsen I had a realization about your [last comment](https://github.com/tomopy/tomopy/pull/375#issuecomment-451009065) while I was playing with this branch just now. I think that the problem that you are trying to address is that `pyctest` runs `nosetests` on the source tree instead of on the installed package. i.e. if `libtomopy.so` isn't copied to the source tree, then the tests fail.

The reason this happens is because when you run tests from *inside* the source directory, the installed TomoPy is overshadowed by the one in the source directory. You need to run the tests from *outside* the source directory for the tests to import the installed TomoPy.

We could solve the problem in a few ways:

1. Copy the tests somewhere else before running them. This is what the `conda-build` environment does.

2. Change the project structure [as described here](https://blog.ionelmc.ro/2014/05/25/python-packaging/#the-structure) such that when you move one folder down into `/tests` the source doesn't overshadow the installed version.

jrmadsen:
> The reason this happens is because when you run tests from _inside_ the source directory, the installed TomoPy is overshadowed by the one in the source directory

Yes, but this does not happen if `tomopy/__init__.py` does not exist (i.e. replaced by `tomopy/__init__.py.in`) because it does not recognize the `tomopy` directory as a package directory and therefore uses the install directory. I don't have any particular issue with doing solution #1 or #2 above though.

carterbox:
Using git describe to automatically change the version number has a few problems:
1. setup.py also needs to be edited
2. The CMake version only accepts three numbered versions, it doesn't like git hashes.
So I'm putting that idea on hold.

prjemian:
Have you considered [versioneer](https://github.com/warner/python-versioneer)?  It uses the Git tag and it is a very easy to configure.  Lots of projects use this.

carterbox:
Thanks for the suggestion @prjemian. In the past, I have used [setuptools_scm](https://github.com/pypa/setuptools_scm), but this problem is challenging because we also want CMake to be aware of the version? I'm actually not sure of @jrmadsen 's motivation for CMake grabbing the version number from `VERSION` besides printing it to stout during the compiling process.

newville:
@carterbox  I second the recommendation of versioneer, especially if the alternative is to require generating `__init__.py` as part of the build. It is often convenient to test out new feature with something like `import tomopy` with uninstalled version running from top-level folder -- this is a simple devel/debug approach and you might come to regret not having readily available (I've made this mistake and recommend against it).

It is also rather common to expect that tests not be run from the top-level folder, but say as `cd test && pytest`.  I think one of the links you gave discussed this too.

jrmadsen:
@carterbox The reason CMake reads VERSION is not particularly needed (beyond the proposed `__init__.py.in template file) and was added as more out of habit -- typically when one builds with CMake, you export your installation to `<PACKAGE_NAME>Config.cmake` and `<PACKAGE_NAME>ConfigVersion.cmake` so that other CMake projects can do `find_package(TomoPy)` when they need to build/link against the package and the version is used such that other projects can do `find_package(TomoPy 1.3.0)` for version requirements. But there does not seem to be much a need for this functionality. 

> It is often convenient to test out new feature with something like `import tomopy` with uninstalled version running from top-level folder -- this is a simple devel/debug approach and you might come to regret not having readily available (I've made this mistake and recommend against it).

@newville I am a bit confused by this. Is something broken or are you having issues? Because I have adopted this approach in other projects (and certainly tested it) without any issues on macOS, Linux, and Windows.

For me, I believe complete separation of the source-tree and build-tree is good practice and `tomopy/__init__.py.in` is one way to do it (you don't create `tomopy/__init__.py` in the source tree unless you do `python setup.py develop` -- `__init__.py` only exists down within `_skbuild` and within the installation) but am fine with whatever method is chosen.

newville:
@jrmadsen  

> > It is often convenient to test out new feature with something like `import tomopy` with uninstalled version running from top-level folder -- this is a simple devel/debug approach and you might come to regret not having readily available (I've made this mistake and recommend against it).

> @newville I am a bit confused by this. Is something broken or are you having issues? Because I have adopted this approach in other projects (and certainly tested it) without any issues on macOS, Linux, and Windows.
> For me, I believe complete separation of the source-tree and build-tree is good practice and `tomopy/__init__.py.in` is one way to do it (you don't create `tomopy/__init__.py` in the source tree unless you do `python setup.py develop` -- `__init__.py` only exists down within `_skbuild` and within the installation) but am fine with whatever method is chosen.

For sure there are many ways to arrange python projects.  In my experience it is both useful and something like "common" or "expected"  to be able to check out a package into a directory (say, tomopy-1.X.Y) and have a folder within that top-level folder that is very close to what the "installed python folder" will look like.  Without loss of generality, let's call this directory "tomopy".    Actually, having it be the installed package name is very common and done with this goal in mind. 

If that "tomopy/" folder has an "__init__.py" file then while sitting in that top-level folder one can do

```
   ~> python
   >>> import tomopy
   >>> print(tomopy.__file__)
   /home/myname/tomopy-1.3.0/tomopy/__init__.py
```

And have everything work with the local (uninstalled!) version of tomopy.  This means that one can alter the python code under the "tomopy/" folder and test things out **without** installing anything.  That can be a very convenient way to work.

Of course, there may need to be some C code compiled, and dlls to find and link.   So not all complex packages do work in this "develop in place" mode, but many smaller projects do (to be clear, tomopy is
not that big).  As it turns out, this works with the code layout in the current master branch:

```
   ~> python setup.py build
   .... <compile, link, put sharedlib into **source** tree> 
   >>> import tomopy
   >>> print(tomopy.__file__)
   /home/myname/tomopy-1.3.0/tomopy/__init__.py
   >>> print(tomopy.util.extern.LIB_TOMOPY)
   <CDLL '/home/tomopy-1.3.0/tomopy/sharedlibs/libtomopy.so', handle ....>
```

One of the keys to this working is that the layout of the source tree will match the layout of the installed package.  Not too surprisingly,  that is how setuptools will normally work.   This very much requires that "tomopy/__init__.py" is there and correct.   Without "tomopy/__init__.py", one has to perform whatever step would be needed to create it or one cannot develop in place.

It is certainly uncommon to generate `__init__.py` from a template as part of the build process.  If the goal is to automatically embed version strings into `__init__.py`  using something like versioneer would be the (much) more common approach.

There is real merit in using "normal python tools" and following "normal python development and packaging" where possible.  Of course, rules are made to be broken, and I have broken many of these myself.  I am usually the last person to invoke "best practice" and I never defend setuptools.   But I've also learned that not following common conventions can be its own punishment.   Breaking this ability to "develop in place" will make working on tomopy harder for some (perhaps even most) developers.   
You want to make it easy for users to become developers.  That is kind of why one uses Python.  Using a template for `__init__.py` will make that harder.

If I thought it would help, I would make a similar plea about adding 2000+ lines of CMake files to tomopy.  Among the things you know about all and future developers and users of tomopy is that they will have some familiarity and some interest in Python.   You do not know if they have familiarity or interest in CMake.

carterbox:
The problems that we are trying address with this discussion are:
1. import parity when testing
2. keeping accessible development for new developers
3. automatic versioning

I *think* I can address all of these concerns by (1) using `versioneer` or `setuptools_scm` because using a CMake file template seems awkward and unnecessary in this case and (2) moving the `tomopy` module into the `src/` directory. In this way, `tomopy/__init__.py` does not need to be altered, and in-place development can be accomplished this way
```bash
$ python setup.py develop
$ cd tests/ && nosetests
```
and out of place development this way
```bash
$ python setup.py install
$ cd tests/ && nosetests
```
As discussed elsewhere, CMake is a common and useful tool for supporting multiple platforms and various compilers which is what we are trying to do.

Thanks for all the helpful input @prjemian, @newville.

newville:
@carterbox Well, again, there are many ways to arrange and work with python packaging and development cycles.

`python setup.py develop` is actually pretty different from working with an uninstalled package in its source directory.   `develop` will redirect the installed version to the local directory, and requires write access.  That can certainly be a useful thing to do (to be clear, I use this often), but it is not the same as working with an uninstalled package, especially if working with a python installation used by multiple processes or accounts or anything "in production".  

I thought this discussion was about using CMake as the build system.   Again, I fully support the tomopy devels working any way they want.  I do hope the cost of using non-standard build tools has been understood fully.

carterbox:
Here's the proposed structure from my [earlier comment](https://github.com/tomopy/tomopy/pull/375#issuecomment-461299001);

```
/benchmarking  # pytest scripts
/binder
/cmake  # CMake modules to autolocate external libraries
/doc
/envs  # requirements files for various platforms
/include  # external facing header files for libtomopy.[so, dll] (installed)
/src
    /libtomopy  # source files and internal headers for libtomopy.[so,dll,pyd]
        /c
        /cuda
        /cxx
        /pybind11
    /tomopy  # python source files (installed)
/test  # python tests
...
/setup.py
```

Using this structure, you can test either the uninstalled package from `/` or the installed package from inside `test/`.

jrmadsen:
Looks good to me except let's hold off on pybind11 for now. Side note from the "CMake project configured for an IDE perspective", this would display slightly better:

```
CMakeLists.txt  # simple top-level CMake script
/source
    CMakeLists.txt  # grit that creates build targets, etc.
    /include  
    /libtomopy  
        /c
        /cuda
        /cxx
    /tomopy  
```

But in terms of importance, it's about a -0.00000001 on a scale of 0 to 10.
This is what it currently looks like because `include` is one level higher than the `CMakeLists.txt` that creates the build targets:
 
screen shot 2019-02-08 at 5 47 13 pm

Oh and since the picture above shows the target, the CMake system has a `format` target so you can just do `python setup.py build -- -- format` (i.e. runs `make format`) to run clang-format.

carterbox:
In my mind, people might want to use libtomopy separately from TomoPy which is why there is a top level folder for the public header. In that case, I should probably reorganize the headers to create a `libtomopy.h` which only contains public functions... but I think no one actually wants that. #YAGNI

Thus, there is no reason why I shouldn't move `/include` to `/src/include`.

PR #376, opened 2018-12-08
Fix to docs failing to build

decarlof:
docs fails to build. See docs [before](https://tomopy.readthedocs.io/en/latest/api/tomopy.misc.corr.html)  and [after](https://tomopy-fdc.readthedocs.io/en/latest/api/tomopy.misc.corr.html#) this pull request is accepted.


carterbox:
@decarlof If you make more versions of the docs available besides `latest` and `stable`, then there is still something to look at when the `latest` version breaks. Also the `stable` docs are still broken. Which version does that point to?

Issue #377, opened 2018-12-10
Proposal: Reduction of dependencies


carterbox:
I propose the following actions in order to reduce the installation footprint of TomoPy:

- [ ] Remove unused dependencies from entire project
- [x] Replace rarely used imports with equivalent functions from frequently used imports
- [x] Separate dependencies which are only used for optional features and testing

Please discuss why or why not these things should be done.

## Unused dependencies
If dependencies are never actually called by functions within the project, they should not be dependencies.

- [ ] Does anyone actually use the FFTW backend anymore? https://github.com/tomopy/tomopy/pull/375#issuecomment-445991938 Probably not. It shall be removed.

## Rarely used imports
Importing an entire library for one function may not be appropriate.
- [x] Drop hard dependency for ```dxchange``` in ```alignment``` and ```recon``` modules. It is used for writing TIFF images only. Use `tifffile` or some other library instead? (#383)
- [x] ~~Drop dependency to ```pywt.dwt2``` and ```pywt.idwt2```  in ```stripe``` module. Use ```from tomopy.util.misc import (fft2, ifft2)``` instead.~~
    - Not valuable since `pywt` is a dependency of `scikit-image`.
-  [x] ~~`scikit-image` is only used for resizing images in `tomopy.misc.phantom`. Perhaps, it's not appropriate to depend on the entirety of `scikit-image` for such a rarely used feature.~~
    - We use the ```register_translation``` in ```skimage.feature``` and ```transform```  in  ```skimage```. The common use is for calculating the rigid registration parameters of two images and auto-centering of the rotation center in tomographic reconstruction.
## Keep testing dependencies separate
Dependencies that are only used for testing should not be included in the final package that is distributed to users.  (e94e4d582503a)

dgursoy:
I think all three points make sense. 

Some notes to remember:
- We use the ```register_translation``` in ```skimage.feature``` and ```transform```  in  ```skimage```. The common use is for calculating the rigid registration parameters of two images and auto-centering of the rotation center in tomographic reconstruction.
- Drop dependency to ```pywt.dwt2``` and ```pywt.idwt2```  in ```stripe``` module. Use ```from tomopy.util.misc import (fft2, ifft2)``` instead.
- Drop dependency to ```dxchange``` in ```alignment``` and ```recon``` modules. It is used for writing TIFF images only.

carterbox:
[...]

PR #383, opened 2018-12-15
Demote dxchange to soft dependency

carterbox:
The main contribution of this pull is to move `dxchange` from required dependency to soft dependency. It also removes some unused import statements and corrects the runtime requirements in meta.yaml and requirements.txt

Addresses some of #377.

This pull does not remove `pyfftw` related code. However, `pyfftw` is not required dependency.

### DIFF
--- a/tomopy/prep/alignment.py
+++ b/tomopy/prep/alignment.py
@@ -81,6 +81,38 @@ __all__ = ['align_seq',
            ]
 
 
+def write_tiff(data, fname='tmp/data', digit=None, ext='tiff'):
+    """
+    Write image data to a tiff file.
+
+    Overwrite existing data and infer data-type from the data.
+
+    Parameters
+    ----------
+    data : ndarray
+        Array data to be saved.
+    fname : str
+        File name to which the data is saved. ``.tiff`` extension
+        will be appended even if it already has one.
+    digit : int
+        Append this number to fname using a dash. e.g. {fname}-{digit}.{ext}
+    """
+    # Add the extension and digit.
+    if digit is None:
+        fname = '{}.{}'.format(fname, ext)
+    else:
+        fname = '{}-{:d}.{}'.format(fname, digit, ext)
+    # Convert to absolute path.
+    fname = os.path.abspath(fname)
+    # Create the directory if it doesn't exist.
+    dname = os.path.dirname(os.path.abspath(fname))
+    if not os.path.exists(dname):
+        os.makedirs(dname)
+    # Save the file.
+    import tifffile
+    tifffile.imsave(fname, data)
+
+
 def align_seq(
         prj, ang, fdir='.', iters=10, pad=(0, 0),
         blur=True, center=None, algorithm='sirt',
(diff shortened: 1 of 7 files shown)


decarlof:
@carterbox makes sense. I will just expand the [docs](https://tomopy.readthedocs.io/en/latest/data.html) to make sure users are aware of  dxchange if they need it.

carterbox:
@decarlof Are you asking me to expand the docs or are you saying that you have already done it?

dgursoy:
We already have this kind of information in the docs. See:
[https://tomopy.readthedocs.io/en/latest/data.html](https://tomopy.readthedocs.io/en/latest/data.html)

dgursoy:
It seems ```tifffile``` is now a dependency. So I'd replace all ```dxchange``` functionality with ```tifffile``` before the merge.

carterbox:
`tifffile` was already a dependency because it is a dependency of `dxchange`.

tomopy/prep/alignment.py
```diff
@@ -339,6 +335,7 @@ def align_joint(
             conv[n] = np.linalg.norm(err)
 
         if save:
+            import dxchange
             dxchange.write_tiff(prj, 'tmp/iters/prj/prj')
```
carterbox: Completely removing `dxchange` requires copying three functions from [`dxchange.writer`](https://dxchange.readthedocs.io/en/latest/_modules/dxchange/writer.html#write_tiff) because this function does stuff like automatically preventing existing files from being overwritten.

decarlof:
I have seen people writing their own readers and missing dxchange perhaps  just changing the section title from 

“Tomographic data files”

to

 “Read Tomographic raw data”

or 
 “How to Read Tomographic data”

will do

francesco

dgursoy:
We don't need to use all functionality of ```dxchange_writer``` such as auto preventing existing files from being overwritten. We can define a TIFF read strategy for the whole package. How about we save files using the name of the function that they were called with a unique numeric index? This way we can use images for debuging or other purposes and their file names won't create any conflict.

carterbox:
I could replace `dxchange.write_tiff()` with the following:

```python
def write_tiff(data, fname='tmp/data', digit=None, ext='tiff'):
    """
    Write image data to a tiff file.

    Overwrite existing data and infer data-type from the data.

    Parameters
    ----------
    data : ndarray
        Array data to be saved.
    fname : str
        File name to which the data is saved. ``.tiff`` extension
        will be appended even if it already has one.
    digit : int
        Append this number to fname using a dash. e.g. {fname}-{digit}.{ext}
    """
    # Add the extension and digit.
    if digit is None:
        fname = '{}.{}'.format(fname, ext)
    else:
        fname = '{}-{:d}.{}'.format(fname, digit, ext)
    # Convert to absolute path.
    fname = os.path.abspath(fname)
    # Create the directory if it doesn't exist.
    dname = os.path.dirname(os.path.abspath(fname))
    if not os.path.exists(dname):
        os.makedirs(dname)
    # Save the file.
    import tifffile
    tifffile.imsave(fname, data)
```

PR #384, opened 2019-01-06
Happy New Year!

decarlof:
updated license


decarlof:
@dgursoy Not sure why is breaking travis. Perhaps is because I keep re-syncing my fork and the original tomopy repository was at one point re-based. I will close this and start from a fresh fork

carterbox:
@decarlof That shouldn't be the reason because you only made changes to the comments in some files. If you look at the reason for the build failure, it says it's a conda dependency resolution problem.

carterbox:
@dgursoy This pull and all current builds are failing because the conda-forge channel stopped building for python 3.5 in [October](https://conda-forge.org/docs/news_announce.html). Our Travis builds for python 3.5 are trying update conda, but it is not compatible with 3.5.

PR #380, opened 2018-12-13
Update find_center_vo

nghia-vo:
Improve the reliability of fine-searching.
Improve SNR by averaging few sinograms
Denoising (smooth filter) is critical. It's part of the method.
Tidy up variable names, codes.

PR #386, opened 2019-01-22
Add ring removal methods. Update docs style

nghia-vo:


nghia-vo:
Is there any chance to merge this PR? There're rooms for improving coding style and debugging, but these algorithms should be ready to be used. They give users more choices to deal with challenging ring artifacts.

carterbox:
@nghia-vo, thanks for your patience. @dgursoy was reviewing your original pull request, but then the holidays and some severe weather disrupted our work here in Chicagoland. I'll remind him about this today.

carterbox:
@nghia-vo This pull request is actually the same as #380, right? You actually don't need to open a new pull request when you make changes. You can just update the old one. Actually, it's probably better if we re-open the old one because all of Doga's comments are there. I'm going to try and do that now.

PR #389, opened 2019-02-12
BUG: Add keyword to suppress skimage API change warning

carterbox:
The skimage API is changing in 0.15, so adding this keyword prevents
the behavior of skimage.transform.resize from changing. This warning
is drowning the tests output on Windows.


carterbox:
Thanks, Jonathan!

PR #390, opened 2019-02-15
NumPy incompatibility and other tweaks

carterbox:
This pull increases the number of parallel jobs on Appveyor and fixes API compatibility issues with skimage and numpy. Please read the commit message for the details. The changes from each commit are orthogonal, so it should be easy to understand what is going on.

### DIFF
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -7,7 +7,7 @@ skip_commits:
    - docs/*
    - '**/*.html'
 skip_branch_with_pr: true
-max_jobs: 1
+max_jobs: 6
 image: Visual Studio 2015
 init:
   - git config --global core.autocrlf input
--- a/.appveyor.yml
+++ b/.appveyor.yml
@@ -1,7 +1,4 @@
 version: 1.0.{build}
-branches:
-  only:
-    - master
 skip_commits:
  files:
    - docs/*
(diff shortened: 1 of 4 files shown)


jrmadsen:
I think you also need to change the `empty_shared_array` function:

```Python
def empty_shared_array(shape, dtype=np.float32):
    # create a shared ndarray with the provided shape and type
    # get ctype from np dtype
    temp_arr = np.empty((1), dtype)
    ctype = type(np.ctypeslib.as_ctypes(temp_arr)._type_())
```

carterbox:
Good catch! I guess I didn't catch this one because empty_shared_array is not called during the tests. I should have searched for all mentions of `_typecodes` in the code base.

dgursoy:
Thanks @carterbox! 

Why do we need multiple parallel jobs? Is it because it'll do all build tests (e.g. for different versions) sequentially if ```max_jobs``` is 1?

carterbox:
That is correct. The build test for each python version is a separate job. When `max_jobs = 1`, each of these independent builds is executed sequentially even though they are independent.

dgursoy:
Got it. Why would anyone want to do it in serial then? I don't clearly see the purpose of setting this parameter manually. It should have been generated automatically based on how many builds that one wants to run. That being said, I'll accept the PR. Thanks!

PR #391, opened 2019-02-15
Simple impulse noise removal and auto background normalization

KedoKudo:
Add two functionalities for tomography reconstruction at APS 1ID:

* selective_median_filter(): remove impulse noise in a single image  

* sino/tomo_normalize_background_aps_1id(): background normalization with auto air region detection.


decarlof:
just one comment/thought: should instrument specific processing functions (like _aps1D) be part of tomopy or go into a beamline specific preprocessing toolbox? On the other end, if these are general and useful for other tomographic instruments we should consider removing the _aps1D and document well when/how to apply (perhaps also adding an example/jupyter notebook in the docs)

KedoKudo:
The appendix pointing towards a specific instrument is mostly to let the user know that the function has been tested at this instrument.  Using it to process data from other instruments is not guaranteed to result in desired results, even if in theory it should.

decarlof:
then these are probably  not general enough to be included in tomopy (at least at this stage) but rather in the dedicated independent toolbox. We kind of had a similar discussion regarding data readers and we decided to pull them out too.

KedoKudo:
That is indeed a good point.  I will close this pull request for now until they are fully tested for other instruments as well.

PR #398, opened 2019-03-12
Updated benchmarks

carterbox:
The main objectives of this pull are as follows:
1. refactor `/benchmarking` to make it more concise
2. use `ms-ssim` as a metric for determining the quality of reconstructions instead of the L2 and L1 norms of the difference between the original and reconstructed images. 
3. make quality ratings as a function of number of iterations for iterative algorithms


benchmarking/__main__.py
```diff
@@ -324,58 +313,52 @@ def run_pyctest():
...
+                    pyctest.PYTHON_EXECUTABLE,
+                    "-m", "benchmarking.phantom",
+                    "-p", phantom,
+                    "-s", "{}".format(nsize),
+                    "-A", "360",
+                    "-f", "jpeg",
+                    "-S", "1",
+                    "-n", "{}".format(args.ncores),
+                    "-i", "{}".format(args.num_iter),
+                    "--output-dir", "Testing",
```
carterbox: Where should the outputs of these tests be saved? Does it matter?

benchmarking/__main__.py
```diff
@@ -188,10 +188,14 @@ def run_pyctest():
...
     pyctest.BUILD_NAME = " ".join(pyctest.BUILD_NAME.split())
     # how to build the code
-    pyctest.BUILD_COMMAND = "{} setup.py install".format(
-        pyctest.PYTHON_EXECUTABLE)
-    # generate the code coverage
+    pyctest.BUILD_COMMAND = " ".join([
+        pyctest.PYTHON_EXECUTABLE, "-m",
+        "pip", "install", pyctest.SOURCE_DIRECTORY, "-vvv",
+        # "--target", pyctest.BINARY_DIRECTORY,
+    ])
```
carterbox: Jonathan says that pip adds noise to the build logs on CMake.



Try the following.

a. Use `python setup.py build -b BINARY_DIRECTORY` so tomopy is installed to a custom location instead of to the site-packages

b. set `WORKING_DIRECTORY` to `BINARY_DIRECTORY/libs` so the benchmarks are executed in a location that can discover tomopy

Thus no annoying messages for you and no annoying installation mess for me.

carterbox:
This branch is no longer up to date with 1.5.x. I will reopen when I have rebased.

PR #404, opened 2019-04-19
1.5.x - GPU support, rotation-based recon, MSVC support

jrmadsen:
- CMake build system to handle addition of code in two new languages: C++ and CUDA
- Python bindings to C++/CUDA code still go through C interface (i.e. no direct binding)
- SIRT and MLEM have been implemented on the GPU and CPU using rotation-based algorithm
    - GPU support has been validated for Windows and Linux
    - CPU version uses OpenCV for rotation
        - OpenCV distributed via conda + MinGW on Windows does not work. Use MSVC compiler on Windows.
    - GPU version uses NPP for rotation
    - benchmarking on NVIDIA P100: ~11x slower than gridrec but vastly improved reconstruction quality
    - benchmarking on NVIDIA V100: per-slice speed-up over ray-based algorithm is ~650x, e.g. a TomoBank reconstruction (2048p + 1,500 proj angles) formerly requiring ~6.5 hours is completed in ~40 seconds
- Support for Microsoft Visual C++ (MSVC) compiler
    - Implemented gridrec in C++ (uses `std::complex`) which is enabled by default on Windows
- To enable new algorithms, include `accelerated=True` to `tomopy.recon` for SIRT and MLEM
    - there are other options available but unless there is an explicitly understanding the effects of the other parameters, use the defaults.
- Multi-GPU support is available
    - Automatic detection of number of available devices
    - Multiple threads started at Python level automatically spread out over the number of available GPUs 
- Secondary thread-pools created in C++ code to provide highly efficient communication with the GPU and additional parallelism on the CPU.
    - When running on the GPU, set `ncore` parameter to `tomopy.recon` to the number of GPUs available. 
    - Each "Python" thread creates a unique secondary thread-pool with a default size of `2 * number-of-cpus`. This is intentional and, in general, the larger the secondary thread-pool, the more efficiently the CPU-GPU communication latency is hidden. However, in general, more than 24 threads per thread-pool provides no benefit (all latency is essentially hidden at that point)

PR #406, opened 2019-04-29
Build with CUDA toolkit on Travis

carterbox:
This purpose of this pull is to add the CUDA toolkit to our build infrastructure so that we can catch CUDA compiling errors earlier.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,11 +9,18 @@ env:
     - PYTHON_VERSION="3.6"
     - PYTHON_VERSION="3.7"
   global:
+    - CUDA_INSTALLER=cuda_10.1.168_418.67_linux.run
+    - CUDA_TOOLKIT_ROOT_DIR=${HOME}/cuda
+    - CUDA_COMPILER=${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc
+    - LD_LIBRARY_PATH=${HOME}/miniconda/lib:${CUDA_TOOLKIT_ROOT_DIR}/lib64
     - PATH=${HOME}/miniconda/bin:${PATH}
-    - LD_LIBRARY_PATH=${HOME}/miniconda/lib
-    - PYCTEST_ARGS='-SF --pyctest-jobs=1 --pyctest-site=Travis --exclude-phantoms shepp3d -- -VV'
+    - PYCTEST_ARGS="-SF --pyctest-jobs=1 --pyctest-site=Travis
+      --exclude-phantoms shepp3d -- -VV --
+      -DCUDA_TOOLKIT_ROOT_DIR=${CUDA_TOOLKIT_ROOT_DIR}
+      -DCMAKE_CUDA_COMPILER=${CUDA_COMPILER}"
     - TOMOPY_USE_C_ALGORITHMS=1
 before_install:
+  # Install miniconda
   - if [[ "$PYTHON_VERSION" == "2.7" ]]; then
       wget https://repo.continuum.io/miniconda/Miniconda2-latest-Linux-x86_64.sh -O miniconda.sh;
     else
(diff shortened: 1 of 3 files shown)


dgursoy:
Is this still on progress? Any explanation about why the tests are failing?

carterbox:
To be clear, merging #404 does not depend on this pull. This pull is something that Jonathan and I had discussed briefly just as an additional sanity check. This pull would catch *compiling errors only* because we do not have a CI with a CUDA GPU set up anywhere.

The current problem that I'm having is that CMake isn't recognizing that the CUDA compiler is installed. It's supposed to check the environment for `CUDA_TOOLKIT_ROOT_DIR`, but it seems to decide that CUDA is not installed even though this file path exists. I suspect one of the following outcomes:

1. I'm not passing this variable to CMake correctly
2. The goals of this pull cannot be accomplished because CMake requires the CUDA drivers be installed, but the CUDA installed will not install drivers without detecting CUDA hardware (which doesn't exist on Travis).

carterbox:
anaconda.org is down. Try again later.

carterbox:
Build is failing with the following message:
```
CMake Error at /home/travis/miniconda/envs/tomopy/share/cmake-3.14/Modules/FindPackageHandleStandardArgs.cmake:137 (message):
  Could NOT find CUDA (missing: CUDA_NVCC_EXECUTABLE CUDA_INCLUDE_DIRS
  CUDA_CUDART_LIBRARY)
```
However, I have passed all of those variables to CMake, so I don't know why it is unhappy. I will need to do some testing locally on a machine sans CUDA GPU.

carterbox:
I think the problem is that CMake doesn't pass the host compiler to `nvcc` when it is testing for a working CUDA compiler, so the search will fail. `nvcc` only searches for `gcc` in the usual locations, and it doesn't take environment variables like `$CC`. I was able to get the CUDA modules to compile with the conda-forge compiler by putting a symbolic link at `/usr/local/bin/gcc`.

carterbox:
One thing to think about is that `1.5.x` has increased Travis build times from ~8 min to ~23 min. Not sure if that's a big deal or not.

jrmadsen:
Compiling more code and building it twice so that's to be expected. Although you don't need the second build since the coverage is CDash got broken.

carterbox:
Now the build time is a problem. Our build times are capped at 50 minutes and the tests are exceeding this allotted time. I'm going separating them into shorter separate builds using the build matrix.

jrmadsen:
Add `--phantom-size 128` like the one in the above. The pyctest script when "coverage" was enabled skipped the built-in phantoms because C/C++ code coverage requires you to compile with zero optimizations so it tends to be very slow. When you removed, the coverage flag, it started reconstructing the built-in phantoms and the default size is 512 pix images.

carterbox:
Thanks for explaining the increased build times. However, I've broken the builds into shorter builds anyways due to the fact that these changes to the C extension shouldn't be affected by the python version. Thus, it's not necessary to build with CUDA or tasking on all of the python versions? Here's the new build matrix:

| python | CUDA | tasking |
| - | - |-|
| 2.7 | False | True |
| 3.6 | False | False |
| 3.6 | False | True |
| 3.7 | False | True |
| 3.7 | True | True |

carterbox:
We're also testing *all* of the 2D phantoms on Travis. Can I cut that back to just one?

jrmadsen:
Build matrix looks fine to me. If you want to scale back the number of phantoms being tested, go for it.

jrmadsen:
Maybe only scale it back to three though. In the event you start seeing weird results somewhere, having 3 tests for each algorithm is the minimum you would need to diagnose whether something is widespread or isolated

jrmadsen:
It looks fine. You can get rid of the TOMOPY_USE_C_ALGORITHMS environment variable. Also, you might have had issues with the conda-forge CUDA bc it wants a clang 6 compiler. I tested it the other day on my Mac. You might want to try adding the LLVM toolchain repo and installing clang via apt-get. I suspect the it's not doing that bc you are installing GCC via conda. Clang tends to produce slower code than GCC so definitely don't make a switch in the env files though,

carterbox:
I'm not using the conda-forge `cudatoolkit-dev` package because it doesn't actually provide the toolkit, it just wraps the installer from NVidia. I'm not sure why they chose to do that, but I found it annoying. I'll remove that `TOMOPY_USE_C` variable in the future.

In terms of building with CUDA for distribution, there is a container with the full cudatoolkit preinstalled [somewhere in development for the conda-forge infrastructure](https://github.com/conda-forge/docker-images/pull/93#issuecomment-497183733). I'm think I'm going to wai for that to arrive because installing the cudatoolkit to the build environment myself is super annoying.

PR #414, opened 2019-06-03
BLD: fixing a tiny typo in setup.py

mrakitin:
The `setup.py` looks really re-worked since I saw it last time (triggered by @carterbox's https://github.com/data-exchange/dxchange/issues/81#issuecomment-498422461), and they are great changes based on scikit-build. I noticed a minor typo in a help message, which is fixed here.


carterbox:
Thanks. This change is included in #406.

mrakitin:
Ups, did not notice that. Thanks for consideration.

PR #420, opened 2019-06-07
Create issue templates

tomopy:
Provides optional templates for users to fill out when requesting help or proposing new features.


carterbox:
@dgursoy, I intentionally chose to not have separate troubleshooting and bug-fix templates because I don't trust most of our users to know the difference between a bug in TomoPy and a bug in their own code. I assume that persons who positively identify a bug don't need a template? We can always add a separate bug-fix template later.

dgursoy:
Makes sense. Thanks again.

PR #429, opened 2019-07-11
Add distortion correction method

nghia-vo:


nghia-vo:
@decarlof , @dgursoy: I've replaced functions based on remap of OpenCV by using map_coordinates of Scipy.  They are quite close in term of speed.

carterbox:
@nghia-vo Do you have some simple tests that we can add to `test/` for our automated testing service?

nghia-vo:
@carterbox : I can do that. If the following steps are enough?
- Copy data to test/test_data
- Write test code in test/test_prep
- Write loaders in util.py (my test data is an image and a text file)
Does the name of the test class need to follow some kinds of conventions? e.g name of class for stripe removal tests is StripeRemovalTestCase

carterbox:
@nghia-vo Thanks! Because I'm not familiar with your algorithms, including tests will help prove that these functions are executing the way you intended.

- Please create a new file `test_prep/test_alignment.py` to match the location of the function that is being tested.
- The name of the class should be formatted CamelCase (Choose a descriptive name).
- The name of the tests in the class *must* start with the word `test_` in order for the test runner to recognize it as a test. If there is one test per function, then the tests may be named `test_function_name`.

dgursoy:
Also please use npy arrays as test cases instead of tiffs. You can create small image sizes for input and output like 16 by 16 or so to minimize the sizes.

nghia-vo:
@carterbox @dgursoy : Looking good?

dgursoy:
Looks good to me! Thanks!

dgursoy:
@carterbox merge?

PR #435, opened 2019-07-23
BLD: Remove license and readme from MANIFEST

### LINKED ISSUE #434: License and README are unpacked in root directory
`LICENSE.txt` and `README.rst` are being unpacked into the root installation directory instead of into the directory `tomopy` or `tomopy.egg-info`. This is the wrong place for these files to be installed and it may cause problems [with packaging](https://github.com/conda/conda-pack/issues/78#issuecomment-513335641).

This may have happened when we moved `tomopy/` from the source root to `source/`.

carterbox:
These files, if included in the MANIFEST, are placed in the root
directory of the python installation instead of with the source. This
can cause problems with packaging because files are placed outside the
tomopy directory. Also, they are not necessary because wheels will
automatically include files named LICENSE* in the dist-info and the
README text can be included with the setup metadata.

Resolves #434

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,2 @@
-include README.rst
-include LICENSE.txt
 graft source/tomopy
 global-exclude *.py[co]
--- a/envs/linux-36.yml
+++ b/envs/linux-36.yml
@@ -25,7 +25,7 @@ dependencies:
   - pywavelets
   - scikit-build
   - scikit-image
-  - scipy<1.3
+  - scipy
   - setuptools_scm
   - setuptools_scm_git_archive
   - six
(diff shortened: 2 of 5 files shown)


tacaswell:
Does this mean the license will not be included in the sdist?

carterbox:
It already wasn't being included for some reason, but it's redundant because the license is included as a header for all of the source files. If you really care about it being included with sdist. I think there may be a way to add the license file using `setup.cfg`?

carterbox:
@tacaswell I couldn't figure out how to get setuptools to add the `LICENSE.txt` file to the package metadata, but I was able to add the contents of the file to the long description of the package.

carterbox:
@jrmadsen This pull also removes the dependency requirement that `scipy<1.3`. Do you remember what that bug was? Is it fixed now?

jrmadsen:
Yea I'm pretty sure that was the issue where their conda feedstock contained openblas symbols that our installations didn't have bc of MKL

PR #438, opened 2019-08-02
 Add Tikhonov regularization with Identity Tikhonov matrix

math-vrn:
A simple extension of the gradient descent method for solving Tikhonov regularization of the form |Rf-g|_2^2+alpha |f-v|_2^2 -> min. Example of usage: 
recon = tomopy.recon(proj, theta, center=rot_center, algorithm='tikh', reg_data=v, reg_par=[-1,alpha],num_iter=32) # where -1 refers to automatic gradient step search


dgursoy:
The test for the added function is failing. @math-vrn could you please double check if the test data, dtypes, etc. are correct?

newville:
@math-vrn  Out of curiosity, how does this compare to using Tikhonov regularization from scikit-learn, as in sklearn.linear_model.Ridge?  

I can believe using custom C code might be faster, but it would also be more code to maintain.   In addition to being well-tested and maintained, one of the great features of `sklearn` is how uniform the API is.  That is,  one could easily try using different regularization penalties (Lasso, LARS, etc).  The scikit-learn docs even have a nice example of tomographic reconstruction (https://scikit-learn.org/stable/auto_examples/applications/plot_tomography_l1_reconstruction.html) using these methods.  The results for the data set they have suggests that L1 regularization (Lasso) is preferable to L2 (Ridge / Tikhonov) in at least some cases....

dgursoy:
I think the simple answer is that we need control on the implementations for performance tuning. Also, we have the freedom to use different data fidelity terms and different solvers if needed (not sure how easy this is in sklearn though).

dgursoy:
Thanks @math-vrn!

newville:
@dgursoy 
> I think the simple answer is that we need control on the implementations for performance tuning. 

Well, is this faster than `sklearn.linear_model.Ridge`?   Maybe the LAPACK etc linear algebra libraries used by `scipy.linalg` and `sklearn` are actually faster and more accurate than hand written single-precision C.  

> Also, we have the freedom to use different data fidelity terms and different solvers if needed (not sure how easy this is in sklearn though).

Indeed, the earlier link suggests that L1 regularization gives better results than L2 regularization.  With `sklearn` (see earlier link), changing from L2 to L1 regularization is one line of python.

carterbox:
@math-vrn Is there a citation that we can add to the docs for this method or is it well known enough that explanation is not needed?

prjemian:
Tikhonov regularization is well-known.  The variation on the method
deserves a literature citation, if available.

PR #441, opened 2019-08-07
Add additional distortion functions to tomopy.sim

carterbox:
Currently `tomopy.sim` only has detector noise models. This pull adds zingers and rings to the distortions. I also plan to add a beam drift simulation, but I haven't completed it yet.


nghia-vo:
@carterbox : "Distortion functions" in your codes are to test pre-processing techniques of cleaning tomographic data, aren't they? This name may cause confusion with the functions of correcting lens distortion.

carterbox:
That's correct; they are for data simulation and testing. I disagree that the name is confusing because these functions are all located in the `sim` ("simulation"). The functions for correcting lens distortion are located in the `corr` ("corrections") module. Both modules have a header in the documentation explaining what functions they contain.

nghia-vo:
Sorry, I didn't mean confusion about using these functions, but about the term "distortion" which may be referred to optics distortion in a detector. As the functions in this module are for simulating artifacts (artifact generators?), we can add optics distortion simulation to this module too.

carterbox:
Could you please explain the difference between the words "distortion" and "artifact" in the x-ray imaging context? e.g. Why is poisson noise an "artifact" while dust on the scintillator is a "distortion"?

nghia-vo:
I never hear dust on the scintillator is a "distortion". It causes ring artifact. Could you send a reference? In X-ray imaging, particularly for X-ray microtomography at synchrotron facilities the detector systems are the combination of a scintillator and a visible optics [system,](https://www.helmholtz-berlin.de/media/media/forschung/funktional/werkstoffe/methods/x_tomo/principle_synchr_xray.png) i.e lens and mirror. Lens can cause radial distortion (barrel distortion or pincushion distortion). Mirror can cause perspective distortion.

carterbox:
I have no references. :smile:  Dust was just an example that I made up. I'm just asking a fundamental question about the difference between the definitions of artifacts and distortions such that I can name the modules in the best possible way.

dgursoy:
Distortion (experimental setup, optical, geometric, etc.) is the cause, and artifact (in final reconstructions) is the effect. I believe they may be interchangeably used depending on the context. E.g. in a simulation module, use of the term distortion (of optics, etc.) makes more sense to better explain what is modeled and in image reconstruction module, use of the term artifact makes more sense.

PR #444, opened 2019-08-15
check the number of processes in get_worker_ncore_slices

malte-storm:
I have noticed that tomopy still initializes many processes even if only a very small number (e.g. 1) of slices is to be reconstructed.
The same would apply to "get_worker_ncore_slices"


carterbox:
Perhaps this would be more readable?

```python 
ncore = min(mp.cpu_count(), axis_size)
```

jrmadsen:
Hold on. I think this is fixed. When I added the GPU stuff I made it such that you don't initialize more threads (TomoPy is single process) than slices.

jrmadsen:
Well I can't seem to find it where I thought it was but come to think of it, I think my patch applied only to when recon was called and not when project was called, which I think this may fix so if it solves that, once the more succinct expression is updated, I approve.

malte-storm:
@carterbox: Your suggestion is more elegant. However, in case the user explicitly specifies the number of cores, your suggestion would be overwritten (if it is in "if not ncore" clause).
A possible fix would be to join it all in 
`ncore = min(mp.cpu_count() if not ncore else ncore, axis_size)`

@jrmadsen: I tested on single slices and tomopy started 16 threads, that is how I noticed.

PR #454, opened 2019-11-26
specify Python 3.6 for binder environment

minrk:
astra-toolbox hasn't made stable builds with Python >3.6 and 3.7 is the default on Binder.

the alternative is to allow dev versions by adding `astra-toolbox/label/dev` channel instead, which does have 3.7 builds.


dgursoy:
Thanks @minrk

PR #457, opened 2020-01-16
Fix astra wrapper for recent astra versions

dmpelt:
This PR fixes the astra wrapper for recent astra versions by using `astra.geom_postalignment` for the center of rotation correction. It also simplifies the wrapper a bit by using the same code for CUDA and CPU algorithms.

Once https://github.com/astra-toolbox/astra-toolbox/issues/226 is resolved, the CPU code can use `astra.geom_postalignment` as well, which would enable sub-pixel shifts for CPU. I'm fine with already merging this before that time and creating a new PR later though.


dgursoy:
Thanks @dmpelt!

Issue #428, opened 2019-07-11
Make all build dependencies optional


carterbox:
This feature is related to the following recent issues with other institutions trying to build TomoPy with the new build system and dependencies:

- https://github.com/NSLS-II/docs/pull/112
- https://github.com/conda-forge/tomopy-feedstock/pull/21
- I also had a request for a no-MKL build (for licensing reasons), but I can't find the Issue.

It seems that people are both not expecting the hard OpenCV and MKL build dependencies, but might not need or want them.

**Describe the solution you would like.**
We should make all of the build dependencies, MKL, OpenCV, and CUDA optional, and return a helpful error if users try to use the algorithms that depend on these libraries which explains that they need to have these libraries installed at build-time.

**Describe alternatives you've considered.**
The alternative is to do nothing.

**Additional context.**
CUDA is already optional at build time, but as described in #422, the CUDA methods will fall back instead of returning an error if CUDA fails.

newville:
@carterbox +1 from me: reducing the dependencies and installing with `pip` would be awesome.

For MKL,  I believe that older versions of gridrec used neither MKL nor FFTW.  It may be possible to have a version that could use fftpack (not the fastest, but included with numpy and scipy as shared libs) if MKL was unavailable.  That would probably be some real work.

carterbox:
I've met a few more people at ICTMS this week that are interested in TomoPy for its preprocessing features, but not reconstruction (because they use fan or come beam). Thus they don't necessarily need the features that MKL and OpenCV provide.

@jrmadsen Do you have time to implement this (Making dependencies optional not replacing MKL FFT)?

jrmadsen:
@carterbox I have made significant progress on a semi-redesign that more-or-less implements a backend scheme for the CUDA/OpenCV implementations. Unfortunately, my GPU clusters are undergoing to maintenance so I won't be able to do full testing until later next week.

Could you handle making MKL optional in the source (and sort out any potential issues)? You can commit directly to the branch on my fork if you want. It should be relatively easy to disable, essentially all you need to do is surround the entire source with the preprocessor:

```cpp
#if defined(TOMOPY_USE_MKL)

// original source...

#else
void gridrec(...)
{
    printf(stderr, "Error! TomoPy did not find MKL during configure time and was thus not compiled "
                  "with MKL support. Without MKL, the gridrec reconstruction algorithm is not "
                  "available. If you would like to be able to utilize gridrec without a dependence "
[...]

PR #458, opened 2020-01-17
Reimplementing gridrec for POWER9

carterbox:
>For now, I started from the master branch and [put back](https://github.com/ssomnath/tomopy/commit/b55da98c2195890b59dde9ad44605d3b1d8322f7) the MKL / FFTW conditions from 1.4.x in `gridrec.c` that were removed in 1.5.0 in my fork.  It is my understanding that I need to look into the installation scripts to include FFTW.

_Originally posted by @ssomnath in https://github.com/tomopy/tomopy/issues/428#issuecomment-575732490_


carterbox:
@ssomnath, I'd like to move the discussion of your implementation for POWER nine to this thread because it is slightly off-topic for #428.

It's good to hear that you were given my previous communications with Lukas which I will partially repeat here for posterity:

> Currently, our c-extensions are loaded from a single shared library using the Python ctypes library. Then they are wrapped with a Python compatibility layer defined in tomopy.util.extern. libtomopy is actually a C library that could be linked to any C/C++ program. The problem with all of our extension functions being compiled together into a single file is that this makes it hard have optional build dependencies because the code is not compiled in a modular way. The main objective of the summer intern will be to break the build process such that multiple shared libraries are created and compiled separately e.g. libtomo-gridrec.so, libtomo-stripe.so, etc. and then move these library into a separate repository. However, you can isolate gridrec without moving it to a separate repository, and the intern will still have plenty to do.
>
>This is the way that I'd like to have the project structured:
>
>```
>/source
>     /libtomo  # eventually moved to separate repository
>         /gridrec  # move gridrec build process and files here
>            /c
>             /cxx
>             /CMakeLists.txt
>         # eventually other algorithms grouped by topic moved here
>     /tomopy
>     /c  # the rest of the stuff remains where it is currently
>     /cxx
>
>    /CMakeLists.txt
>```
>After this isolation is completed, please open a pull request. In this way, any further work that you complete will not affect the rest of the project. If you were thinking of providing an alternative gridrec implementation that uses a different FFT library or adds compiler instructions for POWER9 compilers, I don't have any conditions. I'm not familiar with best practices for separating code that is meant for different compile targets (POWER9, x86, etc) while minimizing duplicate code, but surely our current state could be improved. Whatever you decide to do here, I'll ask @tekinbicer to review.
>
>My understanding is that we have gridrec implemented in both CXX and C but the CXX implementation is only to support the MSVC compiler because there is no cross-platform complex type in C, so it's not necessary to make any changes for both C and CXX unless you want to support Windows.
>
>Please be aware that there are compile time features that were added via CMake modules for the purposes of profiling and debugging by @jrmadsen, but they are undocumented. You may find these useful when you are tuning your implementation for POWER9. Contact Jonathan for more information, and try not to break these performance tools when refactoring.

Please note that in the above communications, I have laid out (1) a request for a specific file structure in order to avoid conflicts with the summer student's project (2) a request for refactoring of the gridrec code for maintainability. Meeting both of those goals would be extremely helpful to our project. :1st_place_medal: 

Please let me know if you need any further help deciphering the project.

ssomnath:
It would be nice if the [installation instructions](https://tomopy-test.readthedocs.io/en/latest/install.html) (especially from source) were updated

carterbox:
@ssomnath, those are not our docs. [These](https://tomopy.readthedocs.io/en/latest/devguide.html) are our docs.

ssomnath:
Oops! Sorry, I pasted wrong link. The one you linked was the one I was intending to mention. Too many open tabs!

I have a question - 
Are all the listed dependencies ([linux 3.7](https://github.com/tomopy/tomopy/blob/master/envs/linux-37.yml) for example) strict? I see some are specific to building, some are for testing, others are for the package itself, etc...

ssomnath:
Few other questions I have:
* I am having a hard time figuring out where ``USE_MKL`` is defined. I grepped the entire source code and could not find where it was defined. Could you point me in the right direction?
* It appears that Cmake is unable to find nvcc. Is there a way I can manually specify the paths for ``CUDA`` etc.?

Below is a list of other features that could not be enabled. Any guidance on which ones are absolutely critical, nice to have and how to manually specify library paths would be great:

-- The following features are NOT defined/enabled (-):
     CMAKE_INSTALL_RPATH_USE_LINK_PATH: Hardcode installation rpath based on link path
     TOMOPY_CXX_GRIDREC: Enable gridrec with C++ std::complex
     TOMOPY_USER_FLAGS: Insert CFLAGS and CXXFLAGS regardless of whether pass check
     TOMOPY_USER_LIBRARIES: Explicit list of libraries to link
     TOMOPY_USE_ARCH: Enable architecture specific flags
     TOMOPY_USE_CLANG_TIDY: Enable clang-tidy (C++ linter)
     TOMOPY_USE_COVERAGE: Enable code coverage for C/C++
     TOMOPY_USE_CUDA: Enable CUDA option for GPU execution
     TOMOPY_USE_GPERF: Enable using Google perftools profiler
     TOMOPY_USE_MKL: Enable MKL support
     TOMOPY_USE_OPENMP: Enable OpenMP (for SIMD -- GNU will enable without this setting)
     TOMOPY_USE_SANITIZER: Enable sanitizer
     TOMOPY_USE_TIMEMORY: Enable TiMemory for timing+memory analysis

carterbox:
1. This [conda build recipe](https://github.com/conda-forge/tomopy-feedstock/blob/master/recipe/meta.yaml) divides the requirements into build, run, and test.

1. Right now, I MKL is required to build TomoPy. The `USE_MKL` flag was added in https://github.com/tomopy/tomopy/pull/375#issuecomment-445936509 before we decided to drop FFTW. Therefore, the `USE_MKL` flag does not actually exist anymore, so I don't know why it is mentioned in the output or why I can't find the line of code that generates that message. @jrmadsen may know something more.

2. When using CMake, you can set your compilers using the following environment variables:
```bash
export CC=clang
export CXX=clang++
export CUDACXX=nvcc
export CUDAHOSTCXX=clang++  # sometimes required when NVCC chooses the wrong host
```
3. That is a list of features that were not enabled **or defined**. The only feature in that list which is required is MKL. The build will fail without MKL. The rest are optional. @jrmadsen could tell you more about how to set up timemory, gperf, and other debugging and performance tools.

jrmadsen:
Yea the `TOMOPY_USE_MKL` is definitely left over from when FFTW was an option.

jrmadsen:
Also, sometimes CMake "doesn't find" CUDA because the check for whether the compiler is valid actually fails. In order words, it finds it but it can't compile a test application so it considers the compiler to be broken. Use `cmake -DCMAKE_CUDA_COMPILER=$(which nvcc) ...` and if cmake actually find nvcc but the compiler check is actually failing, this will cause a CMake error.

ssomnath:
@jrmadsen and @carterbox - Thank you for your guidance.

Here is the [build script](https://github.com/JamesEMcClure/TomographyWorkflow/blob/tomopy_build/build_tomopy.sh) I have been putting together, if you are curious / interested. 

I tried with your feedback. It turns out that FFTW (and many other modules that either support tomopy or its dependencies) available on the target machine (OLCF's Summit) is not compatible with the module (LLVM) that provides clang and clang++. Is there a way to use something else instead of clang and clang++?

I did try to build out of curiosity anyway. The installer is indeed able to find CUDA (not sure about multi GPU support). but (obviously) complains that it is unable to find FFTW which is the alternative to the MKL reconstruction 

-- The following features are defined/enabled (+):
     CMAKE_BUILD_TYPE: Build type (Debug, Release, RelWithDebInfo, MinSizeRel) -- ["Release"]
     CMAKE_CUDA_STANDARD: CUDA STL standard -- ["11"]
     CMAKE_CXX_FLAGS_RELEASE: C++ compiler build-specific flags -- ["-O3 -DNDEBUG"]
     CMAKE_CXX_STANDARD: C++11 STL standard -- ["11"]
     CMAKE_C_FLAGS_RELEASE: C compiler build-specific flags -- ["-O3 -DNDEBUG"]
     CMAKE_C_STANDARD: C languae standard -- ["11"]
     CMAKE_INSTALL_PREFIX: Installation prefix -- ["/gpfs/alpine/stf011/world-shared/mcclure/packages/tomopy/_skbuild/linux-ppc64le-3.6/cmake-install"]
     CUDA_ARCH: CUDA architecture (e.g. '35' means '-arch=sm_35') -- ["35"]
     PTL_CXX_FLAGS: C++ flags for project -- ["-W;-Wall;-Wextra;-Wshadow;-faligned-new;-fopenmp-simd"]
     PTL_C_FLAGS: C flags for project -- ["-W;-Wall;-Wextra;-std=c11;-fopenmp-simd"]
     PYTHON_EXECUTABLE: Python executable (base path is used to locate MKL, OpenCV, etc.) -- ["/gpfs/alpine/stf011/world-shared/mcclure/mcclure-conda/bin/python"]
     PYTHON_INCLUDE_DIR: Python include directory -- ["/gpfs/alpine/stf011/world-shared/mcclure/mcclure-conda/include/python3.6m"]
     PYTHON_LIBRARY: Python library -- ["/gpfs/alpine/stf011/world-shared/mcclure/mcclure-conda/lib/libpython3.6m.so"]
     TOMOPY_USE_CUDA: Enable CUDA option for GPU execution
     TOMOPY_USE_OPENCV: Enable OpenCV for image processing
     TOMOPY_USE_PTL: Enable Parallel Tasking Library (PTL)
     tomopy_CUDA_FLAGS: CUDA NVCC compiler flags -- ["-arch=sm_35;--default-stream;per-thread"]
     tomopy_CXX_FLAGS: C++ compiler flags -- ["-fopenmp-simd;-W;-Wall;-Wextra;-Wno-attributes;-Wno-unused-value;-Wno-unused-variable;-Wno-unknown-pragmas;-Wno-unused-parameter;-Wno-reserved-id-macro;-Wno-implicit-fallthrough;-faligned-new"]
     tomopy_C_FLAGS: C compiler flags -- ["-W;-Wall;-Wextra;-Wno-attributes;-Wno-unused-variable;-Wno-unknown-pragmas;-Wno-unused-parameter;-Wno-reserved-id-macro;-fopenmp-simd"]

-- The following features are NOT defined/enabled (-):
     CMAKE_INSTALL_RPATH_USE_LINK_PATH: Hardcode installation rpath based on link path
     TOMOPY_CXX_GRIDREC: Enable gridrec with C++ std::complex
     TOMOPY_USER_FLAGS: Insert CFLAGS and CXXFLAGS regardless of whether pass check
     TOMOPY_USER_LIBRARIES: Explicit list of libraries to link
     TOMOPY_USE_ARCH: Enable architecture specific flags
     TOMOPY_USE_CLANG_TIDY: Enable clang-tidy (C++ linter)
     TOMOPY_USE_COVERAGE: Enable code coverage for C/C++
     TOMOPY_USE_CUDA_MAX_REGISTER_COUNT: Enable setting maximum register count
     TOMOPY_USE_GPERF: Enable using Google perftools profiler
     TOMOPY_USE_MKL: Enable MKL support
     TOMOPY_USE_NVTX: Enable NVTX for Nsight
     TOMOPY_USE_OPENMP: Enable OpenMP (for SIMD -- GNU will enable without this setting)
     TOMOPY_USE_SANITIZER: Enable sanitizer
     TOMOPY_USE_TIMEMORY: Enable TiMemory for timing+memory analysis

-- Configuring done
-- Generating done
-- Build files have been written to: /gpfs/alpine/stf011/world-shared/mcclure/packages/tomopy/_skbuild/linux-ppc64le-3.6/cmake-build
[2/45] Building C object source/CMakeFiles/libtomopy.dir/c/gridrec.c.o
FAILED: source/CMakeFiles/libtomopy.dir/c/gridrec.c.o 
/sw/summit/llvm/1.0-20190225/bin/clang -DNDEBUG -DTOMOPY_USE_CUDA -DTOMOPY_USE_PTL -Dlibtomopy_EXPORTS -I../../../source/include -I../../../source/PTL/source -isystem /gpfs/alpine/stf011/world-shared/mcclure/mcclure-conda/include -isystem /gpfs/alpine/stf011/world-shared/mcclure/mcclure-conda/include/opencv -O3 -DNDEBUG -fPIC   -W -Wall -Wextra -Wno-attributes -Wno-unused-variable -Wno-unknown-pragmas -Wno-unused-parameter -Wno-reserved-id-macro -fopenmp-simd -std=c11 -std=gnu11 -MD -MT source/CMakeFiles/libtomopy.dir/c/gridrec.c.o -MF source/CMakeFiles/libtomopy.dir/c/gridrec.c.o.d -o source/CMakeFiles/libtomopy.dir/c/gridrec.c.o   -c ../../../source/c/gridrec.c
../../../source/c/gridrec.c:59:14: fatal error: 'fftw3.h' file not found
#    include <fftw3.h>
             ^~~~~~~~~
1 error generated.

carterbox:
> Is there a way to use something else (preferably gcc based) instead of clang and clang++?

Yes! Of course. :laughing: You can set those environment variables to the paths of [most C/C++ compilers](https://cmake.org/cmake/help/latest/manual/cmake-compile-features.7.html#supported-compilers). `clang` was just an example.

In order to compile the gridrec binary with FFTW, you'll need to:

1. Ask CMake to find FFTW using [`find_package()`](https://cmake.org/cmake/help/latest/command/find_package.html). Sometimes, people publish [custom finding modules](https://github.com/egpbos/findFFTW) when the default search strategies don't work well.

2. Then you'll use the [`target_link_libraries()`](https://cmake.org/cmake/help/latest/command/target_link_libraries.html) function to tell CMake that the gridrec binary need to link to FFTW.

As far as I know, we do not have a release which builds with CMake and FFTW; in other words, the last time FFTW was a build option, we were still using Python, bat, and bash scripts. It's probably not a good idea to revisit those days :see_no_evil: .

PR #460, opened 2020-01-18
Create CODE_OF_CONDUCT.md

carterbox:
@dgursoy, I used your email for the contact email on this template because I don't think TomoPy has it's own email address nor does it need one. However, I think I just exposed your email address to the internet. If we keep your email address on this form, you may get spam from people scraping GitHub.


dgursoy:
Use tomopy@googlegroups.com instead?

carterbox:
@dgursoy That email address does not exist!

PR #478, opened 2020-05-12
Added interp_nan function

canismarko:
I use *remove_nan* to get rid of some pesky nan's and inf's that come up from dead/hot pixels, etc. However, the downside is that I have to choose a specific value for all bad values. Instead, I want a function that decides a replacement value for each bad value based on neighboring data.

This PR tests and implements a function that replaces bad data with a local median filter of the surrounding data. It achieves the desired result (mostly) but is most likely horribly inefficient since it calculates the entire median_filtered dataset first and so takes a lot of time and memory. I was hoping to use scipy.interpolate.interpn, but it doesn't really allow for interpolation of dense matrices.

Looking at the rest of the misc/corr.py functions they seem to include some parallelization and memory management. **Are there any simple ways that I can improve this code to make use of some of those optimizations?**

### DIFF
--- a/source/tomopy/misc/corr.py
+++ b/source/tomopy/misc/corr.py
@@ -353,16 +353,22 @@ def remove_neg(arr, val=0., ncore=None):
     return arr
 
 
-def interp_nan(arr, size=3, ncore=None):
-    """
-    Replace NaN values in array by interpolation of neighboring data. 
+def interp_nan(arr, size=(1, 3, 3), ncore=None):
+    """Replace NaN values in array with median of neighboring data.
+
+    The value for *kernel* can be either a scalar, or a tuple with
+    length matching *arr.ndim*. If a scalar is given it will be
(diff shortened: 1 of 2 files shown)


carterbox:
Why not just replace all `nan` with `-np.inf` and let the median filter do its job? If you have a lot of `nan` clustered together, then increase the size of the median filter.

carterbox:
Oh. I guess it's because you only want to change the `nan` values and leave the rest of the data unchanged. In that case, you are trying to do something very similar to `tomopy.misc.corr.remove_outlier()`. It might be best if you modify `remove_outlier` to ensure that it also removes `inf` and (optionally) `nan`. Does that make sense?

canismarko:
@carterbox Yeah, I only want to replace the nan/inf values. I do generally run a conventional median filter on my data first, but sometimes there are nans/infs left over that need to be removed. 

Yeah the ``remove_outlier()`` approach seems like a good way to go without adding too much technical debt. Thanks!

Issue #449, opened 2019-09-19
Modularize reconstruction algorithms using entry points


carterbox:
There has been much discussion about reducing dependency lists recently, and this is because all of the built-in TomoPy algorithms are compiled into a single binary. With this structure, it is impossible to offer TomoPy with less dependencies without rebuilding it. Here, I am proposing to separate all of the compiled (non-python) algorithms into separate _plug-ins_ which could each be built once, packaged, and installed separately in any combination. This approach could also allow us to move the code in the `wrappers` module into the home project of the code being wrapped.

There already exists a built-in mechanism called "entry points" through which a Python package will automatically search for other installed Packages that advertise having an entry point of a certain name. For TomoPy, this could be something called 'tomopy-recon'. We would then break the reconstruction algorithms into separate python packages which all advertise this entry point.

### Additional Reading
https://amir.rachum.com/blog/2017/07/28/python-entry-points/
https://setuptools.readthedocs.io/en/latest/pkg_resources.html#entry-points
https://medium.com/@jherreras/python-microlibs-5be9461ad979

carterbox:
This could also lead to the TomoPy core package becoming `noarch` (platform and architecture agnostic).

newville:
@carterbox  Which reconstruction methods would be available in a python-only version of tomopy?  

Personally, I would be OK with Python + Gridrec.  I might even be OK with Python + Gridrec using just the FFT methods guaranteed to be available with numpy (not MKL or FFTW).  Currently, that is significantly slower than MKL, but my understanding (no direct personal experience) is that the FFT to be used in upcoming numpy is only around a factor of 2 slower than MKL.

carterbox:
@newville None. There are no python-only reconstruction methods, but users that I have met at conferences are only interested in TomoPy for its distortion correction methods because they do not have a parallel beam geometry.

dgursoy:
+1 to Python + Gridrec

PR #483, opened 2020-06-01
Proper modularize

dreycenfoiles:
Folders "misc" and "prep" have been placed within a folder "libtomo" which itself is in the "source" folder. In each of the module folders, there is a CMake file that grabs the proper libraries to build them.


source/libtomo/misc/CMakeLists.txt
```diff
@@ -0,0 +1,56 @@
+
+# Locate sources and headers for this project headers are included for IDEs
+
+file(GLOB tomo_headers
+    ${CMAKE_CURRENT_LIST_DIR}/../../include/*.h
+    ${CMAKE_CURRENT_LIST_DIR}/../../include/*.hh
+    ${CMAKE_CURRENT_LIST_DIR}/../../include/*.hpp
+    ${CMAKE_CURRENT_LIST_DIR}/../../include/*.icc)
```
carterbox, line 8: Is this variable inherited from the top level CMakeList? If so, we shouldn't define it again.

source/CMakeLists.txt
```diff
@@ -70,10 +73,14 @@ endif()
 # create library
 add_library(libtomopy SHARED ${LIBRARY_SOURCES})
 
+
+
+
```
carterbox, line 78: Please remove all extra whitespaces like these.

test/test_tomopy/test_recon/test_rotation.py
```diff
@@ -105,8 +105,8 @@ def test_find_center_vo_with_downsampling(self):
...
-    def test_find_center_pc(self):
-        proj_0 = read_file('projection.npy')
-        proj_180 = image_shift(np.fliplr(proj_0), (0, 18.75), mode='reflect')
-        cen = find_center_pc(proj_0, proj_180)
-        assert_allclose(cen, 73.375, rtol=0.25)
+    # def test_find_center_pc(self):
+    #     proj_0 = read_file('projection.npy')
+    #     proj_180 = image_shift(np.fliplr(proj_0), (0, 18.75), mode='reflect')
+    #     cen = find_center_pc(proj_0, proj_180)
+    #     assert_allclose(cen, 73.375, rtol=0.25)
```
carterbox, line 112: Do not include this change in your pull request. We don't want this test to be disabled upstream.

source/CMakeLists.txt
```diff
@@ -153,6 +163,9 @@ if(NOT SKBUILD)
         WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
 endif()
 
+add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/libtomo/misc)
+add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/libtomo/prep)
+
```
carterbox, line 158: Please move these up to before the python files are copied to binary directory.

source/CMakeLists.txt
```diff
@@ -153,6 +156,7 @@ if(NOT SKBUILD)
         WORKING_DIRECTORY ${PROJECT_BINARY_DIR})
 endif()
 
+
```
carterbox, line 159:
```suggestion
```

source/CMakeLists.txt
jrmadsen: The common method is to do this:

```console
add_subdirectory(libtomo)
```

and then add another CMakeLists.txt in that folder that does:

```console
add_subdirectory(misc)
add_subdirectory(prep)
```

jrmadsen:
Why are two more libraries being built instead of just including that compilation in the regular library?

carterbox:
@jrmadsen, As discussed in #449, I want to break the library into components so they can be distributed separately.

jrmadsen:
Adding extra shared libraries when they aren't used as external link targets (e.g. when you have a code that linking to that library) is generally not a good practice w.r.t. performance -- shared libraries are built as position independent code so all the addresses are w.r.t. a global offset. Thus at runtime, every function call between libraries involves computing that offset. It might not make a difference here but it is something to keep in mind if any of these functions are called _many times_ from `libtomopy` -- I work on another code that sees something like a 20% performance improvement just from using static linking instead of dynamic linking.

jrmadsen:
@carterbox That definitely makes sense for the CUDA and gridrec bits, but keep what I posted above in mind when breaking things up. At the very least, add a compiler flag check for `-flto=thin` and `-flto` and if the compiler supports it, that will enable link-time optimization

jrmadsen:
Use `-flto=thin` if available and `-flto` if that is available and `-flto=thin` is not. I'll skip the details but the former builds faster.

carterbox:
@jrmadsen, Thanks for that insight. :smile: The contents of these shared libraries are completely orthogonal. They will not be linking to each other, so I think performance will be unaffected.

PR #485, opened 2020-06-03
CI: Use VS2017 tools for win build environments

carterbox:
Something is broken when trying to build tomopy on Windows with
the conda VS2015 wrapper nowadays. However, using the newer VS2017
wrapper works. The default build tools for Windows is VS2017 now
anyways? Remember to use -GNinja!


carterbox:
I discovered that all the CI services were disabled when I switched TomoPy from a user to an organization.

PR #486, opened 2020-06-03
MAI: Fix broken CI

carterbox:
I accidentally broke the authentication with our CI services when when we switched the tomopy account to an organization, so the CI hasn't been running for 5 months. The purpose of this pull is to get it back up and running.

# Environment files
Removed the Python 2.7 environments because Python 2 is end of life.

# Travis CI
Remove Python 2 build

# Appveyor
I tried to get this service building again, but I couldn't figure out how to activate a conda environment. I switched to Azure because their support for conda environments is better documented. Additionally, we now explicitly test OSX.


carterbox:
There is a problem where the pyctest benchmarks stall for more than 10 minutes, so Travis fails. However, this failure is seemingly random, so I am going to merge this PR.

PR #491, opened 2020-06-08
CI: Set up self-hosted Azure agent with GPU

carterbox:
Use the self-hosted Docker agent to set up a CI capable of running GPU tests.


carterbox:
OK. I have tests running on an actual GPU, but now I need to provide a reference data for the GPU test.

PR #497, opened 2020-06-25
Made Intel MKL optional

dreycenfoiles:
This makes it that Intel MKL is optional during the build process. By default, MKL will be enabled, but if it is not found, it will be disabled, a warning message will be provided, and the gridrec reconstruction algorithm will be disabled. If the user specifies for MKL to be installed and it is not found, CMake will return an error and the build process will stop.


source/tomopy/util/extern.py
```diff
@@ -126,6 +125,12 @@ def c_shared_lib(lib_name, do_warn=True):
 LIB_TOMOPY_GRIDREC = c_shared_lib("libtomopy-gridrec")
 
 
+def MissingLibrary():
+    raise ModuleNotFoundError(
+        '''Function does not exist due to missing library.
+        Check CMake log for more details.''')
```
carterbox, line 131: This error message is too generic, and triple quotations enforces the line breaks and whitespace.

Instead something like f"The {function}" algorithm is not available because its compiled library is missing. Check whether TomoPy was compiled with the dependencies necessary for this algorithm." not only tells us what the problem is, but it suggests a solution.

PR #498, opened 2020-06-25
Optional OpenCV

dreycenfoiles:
This makes OpenCV optional at build-time. By default, OpenCV is enabled, but if CMake is unable to find OpenCV, it will be disabled, CPU acceleration will also be disabled, and CMake will give a warning. If the user specifies for OpenCV to be used and it is not found, CMake will return an error and the build will stop.


dreycenfoiles:
Splitting files in this way ultimately caused too much code duplication a better method of protecting CPU and GPU functions with Macros was employed.

PR #501, opened 2020-06-30
Fixing Linux_CUDA build on modular-extensions

dreycenfoiles:
This is designed to fix the issues that the modular-extensions branch is having when building with CUDA on Linux


carterbox:
@dreycenfoiles, According to the logs, the CPU versions of the accelerated SIRT and MLEM are failing (not the GPU versions). That's a strange result.

dreycenfoiles:
Fixing the CUDA build only required specifying device='cpu' in the OpenCV accelerated versions of MLEM and SIRT. This change has been directly committed.

carterbox:
Thanks, @dreycenfoiles!

PR #502, opened 2020-07-02
Made MKL dependency optional

dreycenfoiles:
This makes it that Intel MKL is optional during the build process. By default, MKL will be enabled, but if it is not found, it will be disabled, a warning message will be provided, and the gridrec reconstruction algorithm will be disabled. If the user specifies for MKL to be installed and it is not found, CMake will return an error and the build process will stop.


carterbox:
Thanks, @dreycenfoiles!

PR #506, opened 2020-07-08
CI: Use older cuda arch for self-hosted CI

carterbox:
I have old GPUs that I'm using to set up a CI server for TomoPy with CUDA architectures 3.5, 3.5, and 3.0. I observed that building with --cuda-arch=35 breaks the GPU tests on all GPUs while compiling with --cuda-arch=30 allows them all to pass.

Is this the expected behavior? Only one of the GPUs has a 3.0 architecture, so shouldn't the tests pass on two GPUs when --cuda-arch=35?

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -45,7 +45,7 @@ jobs:
     displayName: Create build environment
   - script: |
       source activate tomopy
-      python setup.py install --enable-cuda
+      python setup.py install --enable-cuda --cuda-arch=30
     displayName: Setup and install
   - script: |
       source activate tomopy


jrmadsen:
Sorry, just saw this. CUDA allows you to build for multiple architectures so you can build for arch 30 _and_ 35 and when you run on a device which supports 30, it will execute the 30 code it generated. There is a CMake module which will auto-detect the arch, `<PREFIX>/share/cmake-<VERSION>/Modules/FindCUDA/select_compute_arch.cmake` you might want to look into but be aware that it requires running something on the GPU, which will cause problems for HPC where you might be cross-compiling so you will probably want to do something like:

```cmake
set(TOMOPY_CUDA_ARCH "auto" CACHE STRING "...")

if("${TOMOPY_CUDA_ARCH}" STREQUAL "auto")
    # use select_compute_arch.cmake
else()
    # set explicitly 
endif()
```

carterbox:
Hmm... Are there significant performance benefits to compiling for a specific architecture instead of just letting NVCC JIT compile the PTX file? Like should be we be targeting newer architectures even if we don't use features from those architectures?

Related: https://docs.nvidia.com/cuda/ampere-compatibility-guide/#application-compatibility-on-ampere

PR #507, opened 2020-07-09
BLD: Always test cuda compiler if CUDA is found

carterbox:
Something that has bothered me for a very long time is that CMake's `find_package(CUDA)` will succeed, but then `check_language(CUDA)` will fail without checking `${CUDA_TOOLKIT_ROOT_DIR}/bin/nvcc`. With this PR, if CUDA is located and the user has not explicitly disabled CUDA, then CMake will always check whether the CUDA compiler works (and raise an error if it doesn't).


jrmadsen:
You might want to just disable using `find_package(CUDA)` (which I believe is necessary for the libraries) until you are certain that NVCC works. Here is how I do it these days: https://github.com/NERSC/timemory/blob/7d8650277b86a1e56dd27088fb5694e734c468b0/cmake/Modules/Options.cmake#L62

carterbox:
I can't be certain that NVCC works unless I test it, but `check_language(CUDA)` doesn't do its job and actually find NVCC! It would be great if `check_language` took hints, but it doesn't.

PR #511, opened 2020-07-21
Extern Submodules

dreycenfoiles:
This PR continues the trend of modularizing TomoPy's libraries. The file extern.py is now a folder that contains files, one for each of the share object library file that can compiled during the build process (accel, gridrec, misc, prep, and recon). All of the files in the new extern folder only have functions that call from their respective shared library. None of these changes affect the front-end Python API.


source/tomopy/util/__init__.py
```diff
@@ -48,3 +48,38 @@
...
+    if do_warn is True:
+        logger.warning('OSError: ' +
+                       'The following shared lib is missing!\n{}'.format(
+                           sharedlib))
+
+
+def MissingLibrary(function):
+    print(f"The {function} algorithm is unavailable."
+          " Check CMake logs to determine if TomoPy was"
+          " built with dependencies required by this algorithm.")
```
carterbox, line 85: Why are these defined in `util.__init__` instead of `extern.__init__`? Do other modules in `util` use these functions? If not, please move these to into the `extern` module.

source/tomopy/util/extern/gridrec.py
```diff
@@ -0,0 +1,102 @@
...
+
+import sys
+import os.path
+import ctypes
+import numpy as np
+import tomopy.util.dtype as dtype
+import logging
+import warnings
+from .. import c_shared_lib
+from .. import MissingLibrary
```
carterbox, line 64: For each of the new submodules, please remove unused imports. Then arrange the remaining imports into three groups:
```
import standard_library_modules
# whitespace
import third_party_modules
# whitespace
import local libraries
```
Sort alphabetically within each group.

We also do not need `from __future__` imports in these new modules anymore. Those imports were for backwards compatibility with python 2 https://docs.python.org/3.8/library/__future__.html

source/tomopy/util/__init__.py
dreycenfoiles: No other modules in util use the function. However, if they are placed in extern.__init__, it causes a problem of circular dependencies because extern.__init__ is importing accel, gridrec, prep, misc, and recon, but those same files would have to then import extern.__init__ to access the above functions. I've tried changing the order of import to no success. The only solution I've found is to place the above functions in util.__init__, which gets rid of the circular dependency.

carterbox: Which version of Python are you using?

dreycenfoiles: Python 3.7.6.

dreycenfoiles: Python 3.7.6.

carterbox: Works for me on python 3.7.6 and 3.6.10.

dreycenfoiles: I see. I think my issue was that I wasn't telling the files the correct place to get the c_shared_lib function.

PR #512, opened 2020-07-23
Modular extensions

### LINKED ISSUE #428: Make all build dependencies optional

### LINKED ISSUE #448: Minimal dependencies to build tomopy for reconstruction
**Describe the problem**
We are looking at the best way of adding TomoPy to Tomviz.  We have a Python environment with Python 3.7, SciPy, NumPy, and have a PR to add h5py. The dependency set looked larger than expected, including OpenCV and several other components. What is the minimal set of dependencies/build configuration to support reconstruction?

**Expected behavior (if applicable)**
Ideally no big new dependencies, perhaps we need to understand how to turn off optional features.

**Helpful images (if applicable)**
Add screenshots to help explain your problem.

**Platform Information:**
 - OS: We want o support Linux, macOS, and Windows
 - Python Version 3.7
 - TomoPy Version Latest/master

**Additional context**
This is really us exploring  packaging, another approach is to use a Docker container with a miniconda starting point.

### LINKED ISSUE #452: 2020.04.26 Intern - SULI
I am [writing a project proposal](https://docs.google.com/document/d/1fYKdYe_87yAkXHk-QgIB0YxkD3v6H1d4V29qO-LakZc/edit?usp=sharing) for an undergraduate intern to refactor TomoPy. My current proposal addresses the following issues: #428, #449, #448, #422

Since, this is a very public project, I am posting the proposal for public feedback, and so no one is surprised in a few months when code is moved around. Basically, if you have an idle branch that you are planning to merge, please open a pull-request before **March 1, 2020**. This project may include API breaking changes in the Python layer, and it will definitely have API breaking changes for the C layer.

If anyone knows any undergrads who may be particularly suited to this project, please contact me via email. My contact information is available on the Argonne National Laboratory public website. I'm not going to link it directly because I don't want spam.

dreycenfoiles:
Previously, TomoPy has had the required dependencies OpenCV and MKL. These libraries provide CPU acceleration and the gridrec reconstruction algorithm, respectively. However, users of TomoPy may not have these dependencies installed and may not even need the functionality provided by them, so it does not make sense for them to be required during the TomoPy build process.

The best way to make the dependencies optional was to modularize TomoPy.  This is accomplished by breaking up the shared-object library file that was originally created into five distinct library files that have separate dependencies from each other. These libraries are "accel" (the accelerated MLEM and SIRT functions), "gridrec" (the gridrec reconstruction algorithm), as well as "misc", "prep" and "recon", which contain the C implementations of functions required in the Python files under the same directory titles. Under this new file organization, if MKL or OpenCV are not found, the corresponding shared-library object file will not be able to build, but the other library files still will. 

This PR also has organized the header files so that they are now divided into one of five folders depending on which library they provide forward declarations for.

It has also turned `extern.py` into a module with each of the five library files being one of its submodules where the corresponding library file is opened by Python. 

This removed functions from the headers that are not meant to be in the public API, since it was not possible to remove all non-public functions from the external headers, internal headers were created to provide forward declarations when multiple files rely on a function. 

Closes #428, closes  #448. closes #452


cmake/Modules/BuildSettings.cmake
```diff
@@ -18,7 +18,7 @@ set(CMAKE_CXX_COMPILE_FEATURES
 #
 set(CMAKE_INSTALL_MESSAGE LAZY)
 set(CMAKE_C_STANDARD 11 CACHE STRING "C language standard")
-set(CMAKE_CXX_STANDARD 11 CACHE STRING "CXX language standard")
+set(CMAKE_CXX_STANDARD 14 CACHE STRING "CXX language standard")
```
carterbox, line 21: Why is this change necessary?

cmake/Modules/Packages.cmake
```diff
@@ -40,31 +40,39 @@ if(PYTHON_EXECUTABLE)
         ${PYTHON_ROOT_DIR}/include)
 endif()
 
-
 ################################################################################
 #
-#        MKL (required)
+#        OpenCV (required for CPU acceleration)
 #
 ################################################################################
```
carterbox, line 47: This OpenCV and MKL section were swapped in your branch. Reswap them so that the git diff is cleaner (to avoid unnecessary changes upstream).

source/libtomo/gridrec/CMakeLists.txt
```diff
@@ -0,0 +1,59 @@
...
+    PREFIX ""
+    RUNTIME_OUTPUT_DIRECTORY    ${OUTPUT_PATH}/tomopy/util/extern
+    LIBRARY_OUTPUT_DIRECTORY    ${OUTPUT_PATH}/tomopy/util/extern
+    ARCHIVE_OUTPUT_DIRECTORY    ${OUTPUT_PATH}/tomopy/util/extern
+    INCLUDE_DIRECTORIES         "${TARGET_INCLUDE_DIRECTORIES}"
+    C_STANDARD                  ${CMAKE_C_STANDARD}
+    C_STANDARD_REQUIRED         ${CMAKE_C_STANDARD_REQUIRED}
+    CXX_STANDARD                ${CMAKE_CXX_STANDARD}
+    CXX_STANDARD_REQUIRED       ${CMAKE_CXX_STANDARD_REQUIRED}
+    ${CUDA_PROPERTIES})
```
carterbox, line 47: For these new CMakeLists, let's clean them up so they aren't creating unnecessary links. For example, gridrec does not need `${CUDA_PROPERTIES}` and most of the libraries do not need to set a `CXX_STANDARD`.

source/libtomo/misc/CMakeLists.txt
```diff
@@ -0,0 +1,51 @@
...
+
+# core sources
+set(MISC_LIBRARY_SOURCES ${misc_tomo_headers} ${misc_tomo_sources})
+
+# create library
+add_library(libtomopy-misc SHARED ${MISC_LIBRARY_SOURCES})
+
+# link library
+target_link_libraries(libtomopy-misc PUBLIC ${EXTERNAL_LIBRARIES}
+    PRIVATE ${EXTERNAL_PRIVATE_LIBRARIES})
```
carterbox, line 20: Libraries like misc do not have any external dependencies, so they should not be linked against `${EXTERNAL_LIBRARIES}`.

cmake/Modules/BuildSettings.cmake
dreycenfoiles: We were running into segfaults and you found that it was fixed by setting the standard to 14. This was what you told me:

"I was playing with your build, and I observed that both cxx and c gridrec work for Linux if the CXX_STANDARD is set correctly (it must be at least 14 to enable std::complex). The accelerated functions cause a segmentation fault."

carterbox: Hmmm... I don't trust past me. Let's recheck this.

dreycenfoiles:
> Libraries like misc do not have any external dependencies, so they should not be linked against `${EXTERNAL_LIBRARIES}`. 

When I remove that line and compile, I receive errors about undefined functions like `downsample`, `upsample`, `pow`, and `sincos`.

source/libtomo/misc/CMakeLists.txt
dreycenfoiles: When I remove the call to `${EXTERNAL_LIBRARIES}` and build, I get an error saying I have undefined functions liked `downsample`, `upsample`, `atan2`, and `sincos`.

cmake/Modules/BuildSettings.cmake
dreycenfoiles: I think past you was right. When building on Windows, I run into errors when the CXX_STANDARD is 11 when building gridrec. Setting the standard to 14 fixes the issue.

source/libtomo/gridrec/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+    CXX_STANDARD                ${CMAKE_CXX_STANDARD}
+    CXX_STANDARD_REQUIRED       ${CMAKE_CXX_STANDARD_REQUIRED})
+
+target_compile_definitions(libtomopy-gridrec PUBLIC
+    ${${PROJECT_NAME}_DEFINITIONS})
+
+target_compile_options(libtomopy-gridrec PUBLIC
+    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CXX>:${${PROJECT_NAME}_CXX_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CUDA>:${${PROJECT_NAME}_CUDA_FLAGS}>)
```
carterbox, line 54:
```suggestion
    $<$<COMPILE_LANGUAGE:CXX>:${${PROJECT_NAME}_CXX_FLAGS}>)
```

source/libtomo/misc/CMakeLists.txt
```diff
@@ -0,0 +1,49 @@
...
+    C_STANDARD                  ${CMAKE_C_STANDARD}
+    C_STANDARD_REQUIRED         ${CMAKE_C_STANDARD_REQUIRED})
+
+target_compile_definitions(libtomopy-misc PUBLIC
+    ${${PROJECT_NAME}_DEFINITIONS})
+
+target_compile_options(libtomopy-misc PUBLIC
+    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CXX>:${${PROJECT_NAME}_CXX_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CUDA>:${${PROJECT_NAME}_CUDA_FLAGS}>)
```
carterbox, line 45:
```suggestion
target_compile_options(libtomopy-misc PUBLIC
    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>)
```

source/libtomo/prep/CMakeLists.txt
```diff
@@ -0,0 +1,36 @@
...
+    C_STANDARD                  ${CMAKE_C_STANDARD}
+    C_STANDARD_REQUIRED         ${CMAKE_C_STANDARD_REQUIRED})
+
+target_compile_definitions(libtomopy-prep PUBLIC
+    ${${PROJECT_NAME}_DEFINITIONS})
+
+target_compile_options(libtomopy-prep PUBLIC
+    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CXX>:${${PROJECT_NAME}_CXX_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CUDA>:${${PROJECT_NAME}_CUDA_FLAGS}>)
```
carterbox, line 33:
```suggestion
target_compile_options(libtomopy-prep PUBLIC
    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>)
```

source/libtomo/recon/CMakeLists.txt
```diff
@@ -0,0 +1,38 @@
...
+    target_link_libraries(libtomopy-recon PUBLIC m)
+endif()
+
+target_compile_definitions(libtomopy-recon PUBLIC
+    ${${PROJECT_NAME}_DEFINITIONS})
+
+target_compile_options(libtomopy-recon PUBLIC
+    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CXX>:${${PROJECT_NAME}_CXX_FLAGS}>
+    $<$<COMPILE_LANGUAGE:CUDA>:${${PROJECT_NAME}_CUDA_FLAGS}>)
```
carterbox, line 34:
```suggestion
target_compile_options(libtomopy-recon PUBLIC
    $<$<COMPILE_LANGUAGE:C>:${${PROJECT_NAME}_C_FLAGS}>)
```

source/libtomo/gridrec/CMakeLists.txt
carterbox: Double check this. I found some places in the suggestions below where CUDA properties were being set on libraries that did not use CUDA.

PR #513, opened 2020-07-24
Clean headers

dreycenfoiles:
This PR removes non-public functions from header files in the `include` directory. All functions that are not called externally are moved to their corresponding folder under `source/libtomo`.


source/include/gridrec/filters.h
```diff
@@ -58,12 +58,6 @@
 #endif
 #define ANSI
```
carterbox, line 59: These compiler defs and include statements should not be in the public headers because they are unused here. If they are implementation related, they should be in the source files.

source/libtomo/gridrec/tables.h
```diff
@@ -50,11 +50,9 @@
 // Use X/Open-7, where posix_memalign is introduced
 #define _XOPEN_SOURCE 700
 
-#include "filters.h"
 #include "mkl.h"
```
carterbox, line 53: Why does tables.h need mkl? Double check all these gridrec module headers. I think only gridrec.c and gridrec.cc need mkl.

source/include/gridrec/filters.h
```diff
@@ -41,70 +41,139 @@
...
 // POSSIBILITY OF SUCH DAMAGE.
 
-#pragma once
+// Possible speedups:
+//   * Profile code and check adding SIMD to various functions (from OpenMP)
 
-#include <complex.h>
-#include <stdlib.h>
+//#define WRITE_FILES
+#define _USE_MATH_DEFINES
```
carterbox, line 48:
```suggestion
#ifndef M_PI
#    define M_PI 3.14159265359
#endif
```
I think we are defining M_PI directly because not every compiler has `_USE_MATH_DEFINES`.

carterbox:
Thanks, @dreycenfoiles! 🎆

PR #517, opened 2020-09-03
scikit-image API update

s-sajid-ali:
The function `feature.register_translation` is deprecated and users are to move to `registration.phase_cross_correlation`. The conda recipe might also need to set `0.17` as the minimum version of `scikit-image` (which it already does).

### DIFF
--- a/envs/linux-36.yml
+++ b/envs/linux-36.yml
@@ -24,7 +24,7 @@ dependencies:
   - python-coveralls
   - pywavelets
   - scikit-build
-  - scikit-image
+  - scikit-image>=0.17
   - scipy
   - setuptools_scm
   - setuptools_scm_git_archive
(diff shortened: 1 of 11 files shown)


carterbox (changes requested):
Hi @s-sajid-ali,

Thanks for bringing this to our attention. I did notice deprecation warning a few months ago, but I think this was before `scikit-image=0.17` was released. Could you please add the `scikit-image>=0.17` version requirement to the environment files in `/envs`? This will ensure that our CI bots are testing your changes against the correct version. I will handle updating the requirements in our conda-forge feedstock once I figure out why it is not building against the current pinnings.

Thanks again!

carterbox (approved):

carterbox:
Thanks, @s-sajid-ali!

PR #520, opened 2020-09-17
DOC: Make it clearer which shared libraries are optional

carterbox:
Have TomoPy raises errors when loading required components and
only raise warnings when looking for optional components so that it
is clearer to the user which components are considered required
and which are optional.

Regardless, an error is now raised when a user tries to run code
whose shared library is missing instead of just failing silently
and giving bad results.


dreycenfoiles (approved):

carterbox:
Thanks for the review, @dreycenfoiles!

PR #521, opened 2020-09-17
Add tests for prep.stripe, add an interpolation-based method, update parameters of methods in stripe.py

nghia-vo:


carterbox (commented):
@nghia-vo, thanks for this PR!

I see that some function arguments are now required instead of optional (you removed some default values then put recommended values in the docs). Doing this may break existing programs written by others. Is there a specific reason why removing/changing the default values is necessary? Or can we keep the previous defaults and also have new features.

nghia-vo:
@carterbox : These default values are suitable for X-ray tomographic data and for projection-image with size ~ 2k x 2k. They may be not suitable for other types of tomographic data. But I agree the updates may break old codes.  Will bring back the default values and update the PR.

nghia-vo:
I've updated the PR, also fixed a zero division bug #522

source/tomopy/prep/stripe.py
```diff
@@ -679,6 +679,7 @@ def _detect_stripe(listdata, snr):
         xlist[ndrop:-ndrop - 1], listsorted[ndrop:-ndrop - 1], 1)
     numt1 = _intercept + _slope * xlist[-1]
     noiselevel = np.abs(numt1 - _intercept)
+    noiselevel = np.clip(noiselevel, 1e-6, None)
```
carterbox, line 682: Is this clipping operation related to the zero-division?

source/tomopy/prep/stripe.py
```diff
@@ -687,30 +692,31 @@ def _detect_stripe(listdata, snr):
...
-    list1 = np.mean(sinosorted[ndrop:nrow - ndrop], axis=0)
-    list2 = np.mean(sinosmoothed[ndrop:nrow - ndrop], axis=0)
-    listfact = list1 / list2
+    ndrop = int(0.5 * drop_ratio * nrow)
+    sinosort = np.sort(sinogram, axis=0)
+    sinosmooth = median_filter(sinosort, (1, size))
+    list1 = np.mean(sinosort[ndrop:nrow - ndrop], axis=0)
+    list2 = np.mean(sinosmooth[ndrop:nrow - ndrop], axis=0)
+    listfact = np.divide(list1, list2,
+                         out=np.ones_like(list1), where=list2 != 0)
```
carterbox, line 707: Why is it correct to replace division by zero with 1 in this algorithm? Is it because we expect the two averages to be approximately equal?

source/tomopy/prep/stripe.py
```diff
@@ -351,19 +352,20 @@ def remove_stripe_sf(tomo, size=5, ncore=None, nchunk=None):
     return arr
 
 
-def remove_stripe_based_sorting(tomo, size=None, ncore=None, nchunk=None):
+def remove_stripe_based_sorting(tomo, size=21, dim=1, ncore=None, nchunk=None):
```
carterbox, line 355: This new default size is not equivalent to the old behavior. The old behavior is approximately:
```python
size = max(5, min(int(0.01 * tomo.shape[2]), 21)) if size is None else size
```
Is there a reason that `21` is the preferred filter size for all tomogram sizes?

carterbox (changes requested):
@nghia-vo, thanks for addressing that zero-division error in addition to adding the new stripe suppression method. I have some questions below:

source/tomopy/prep/stripe.py
nghia-vo: Yes, it is. This is a precaution for the case that the input is noise-free (e.g simulation data).

source/tomopy/prep/stripe.py
nghia-vo: Correct. This is a normalization step, so the ratios are around 1. Explanation is shown here: https://sarepy.readthedocs.io/toc/section3_1/section3_1_3.html

source/tomopy/prep/stripe.py
nghia-vo: Yeap, it's not a general number. Will roll back.

carterbox (approved):

carterbox:
Thanks, @nghia-vo! :smile:

PR #523, opened 2020-09-30
Enable parallel computing for find_center_vo, allow 2D sinogram input…

nghia-vo:
…, fix bug caused by downsampling

### DIFF
--- a/source/tomopy/recon/rotation.py
+++ b/source/tomopy/recon/rotation.py
@@ -306,9 +306,11 @@ def _search_coarse(sino, smin, smax, ratio, drop, ncore=None):
     Coarse search for finding the rotation center.
     """
     (nrow, ncol) = sino.shape
-    start_cor = ncol//2 + smin
-    stop_cor = ncol//2 + smax
     cen_fliplr = (ncol - 1.0) / 2.0
+    smin = np.int16(np.clip(smin + cen_fliplr, 0, ncol - 1) - cen_fliplr)
+    smax = np.int16(np.clip(smax + cen_fliplr, 0, ncol - 1) - cen_fliplr)
+    start_cor = ncol // 2 + smin
+    stop_cor = ncol // 2 + smax
     flip_sino = np.fliplr(sino)
     comp_sino = np.flipud(sino)  # Used to avoid local minima
     list_cor = np.arange(start_cor, stop_cor + 0.5, 0.5)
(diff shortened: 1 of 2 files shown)


source/tomopy/recon/rotation.py
```diff
@@ -203,14 +204,14 @@ def _find_center_cost(
 
 
 def find_center_vo(tomo, ind=None, smin=-50, smax=50, srad=6, step=0.25,
-                   ratio=0.5, drop=20):
+                   ratio=0.5, drop=20, parallel=True):
```
carterbox, line 207: Please match the rest of the library by using `ncore, nchunk` parameters with default value of `None`. The `distribute_jobs` function automatically chooses values for these parameters.

https://github.com/tomopy/tomopy/blob/1047c0b59de3bf5026209b670ba4b1ebd03ba9f9/source/tomopy/prep/stripe.py#L88-L90

carterbox, line 207: Please match the rest of the library by using `ncore, nchunk` parameters with default value of `None`. The `distribute_jobs` function automatically chooses values for these parameters.

https://github.com/tomopy/tomopy/blob/1047c0b59de3bf5026209b670ba4b1ebd03ba9f9/source/tomopy/prep/stripe.py#L88-L90

source/tomopy/recon/rotation.py
```diff
@@ -250,49 +256,111 @@ def find_center_vo(tomo, ind=None, smin=-50, smax=50, srad=6, step=0.25,
...
-        fft2sino = np.fft.fftshift(fft2(sino_sino))
-        np.abs(fft2sino, out=abs_fft2_sino)
-        abs_fft2_sino *= mask
-        list_metric[i - smin] = abs_fft2_sino.mean()
+    if parallel is True:
+        list_metric = distribute_jobs(np.float32(list_shift),
+                                      _calculate_metric, axis=0,
+                                      args=(sino, flip_sino, comp_sino, mask),
+                                      ncore=mp.cpu_count() - 1, nchunk=1)
+    else:
```
carterbox, line 349: For example, instead of setting `parallel=False`, users can set `ncore=1` and `nchunk=1`.

carterbox (changes requested):
Hi @nghia-vo. thanks for the PR. In addition to the comments below, could you please separate the changes for the bug fix and the "allowing of 2D inputs" into their own commits? It will help me understand which changes go with which objective.

nghia-vo:
Update and separate commits as requested. However nchunk must be set to 1 for the function to work.

carterbox:
I think it's fine if `nchunk=1` is hardcoded in this case. I also removed the `if ncore !=1 else` statement. The number of cores should not affect the result of `distribute_jobs()`.

nghia-vo:
I've checked the update. Good to go.

PR #530, opened 2021-03-01
ENH: Tune use of CUDA NPP for multiple-GPUs

jrmadsen:
@dgursoy @carterbox I getting [NESAP](https://www.nersc.gov/research-and-development/nesap/) benchmarking number for our test-bed that resembles Perlmutter and encountered this issue with tomopy not effectively using multiple GPUs. I feel like we uncovered this a long time ago but resolving it got lost when I was transitioning to my new position.

Here is the FOM values I have:

|   | Algorithm | Filename | Cores | Iterations | # Slices | Dimensions | Time to Solution (sec) |
|--| ---------- | --------- | ----- | ---------- | ------- | ------------ | ---------------------- | 
| Edison | sirt | tomo_0001 | 48 | 100 | 24 | 2048 x 2048 | 28069 |
| Edison | mlem | | | | |  | 28733 |
| Perlmutter | sirt (NN) | | 32 | | |  | 569 |
| Perlmutter | mlem (NN) | | | | |  | 566 |

I only used 32 threads (4 threads per GPU) of the 256 available bc you get pretty much the same performance anywhere for 2+ threads per-GPU (there is a slight boost for 2 threads vs. 1 thread) -- it appears NPP holds a lot of state internally and thus isn't great for parallel execution with multiple streams/threads. If one slice is reconstructed at a time, instead of 3 like in this test, the reconstruction time drops down to < 1.5 seconds, however with 3 slices the reconstruction time is ~5.5 seconds instead of 1.5 * 3 = 4.5 seconds. Thus, there appears to be some speed-up missing.

### DIFF
--- a/source/libtomo/accel/gpu/common.cu
+++ b/source/libtomo/accel/gpu/common.cu
@@ -38,9 +38,9 @@
 //   TOMOPY CUDA implementation
 
 #include "common.hh"
+#include "cuda.h"
 #include "macros.hh"
 #include "utils.hh"
-#include "cuda.h"
 
 //======================================================================================//
 
@@ -284,8 +284,15 @@ cuda_device_query()
     else
         printf("Detected %d CUDA capable devices\n", deviceCount);
 
+    int specific_device = GetEnv("TOMOPY_DEVICE_NUM", -1);
+
     for(int dev = 0; dev < deviceCount; ++dev)
     {
+        // if the process set a specific device, only print out info for that device
+        if(specific_device >= 0 && specific_device < deviceCount &&
+           dev != specific_device)
+            continue;
+
         cudaSetDevice(dev);
         cudaDeviceProp deviceProp;
         cudaGetDeviceProperties(&deviceProp, dev);
--- a/source/libtomo/accel/gpu/mlem.cu
+++ b/source/libtomo/accel/gpu/mlem.cu
@@ -175,6 +175,7 @@ mlem_cuda(const float* cpu_data, int dy, int dt, int dx, const float*, const flo
     // compute some properties (expected python threads, max threads, device assignment)
     int pythread_num = ntid++;
     int device       = pythread_num % cuda_device_count();  // assign to device
+    device           = GetEnv("TOMOPY_DEVICE_NUM", device);
 
     TIMEMORY_AUTO_TIMER("");
 
--- a/source/libtomo/accel/gpu/mlem.cu
+++ b/source/libtomo/accel/gpu/mlem.cu
@@ -124,6 +124,7 @@ mlem_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
 
     // synchronize the stream (do this frequently to avoid backlog)
     stream_sync(stream);
+    CUDA_FAST_CHECK_LAST_ERROR();
 
     // reset destination arrays (NECESSARY! or will cause NaNs)
     // only do once bc for same theta, same pixels get overwritten
@@ -156,7 +157,7 @@ mlem_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
         CUDA_CHECK_LAST_STREAM_ERROR(stream);
 
         // synchronize the stream (do this frequently to avoid backlog)
-        stream_sync(stream);
+        // stream_sync(stream);
     }
 }
 
@@ -175,7 +176,7 @@ mlem_cuda(const float* cpu_data, int dy, int dt, int dx, const float*, const flo
     // compute some properties (expected python threads, max threads, device assignment)
     int pythread_num = ntid++;
     int device       = pythread_num % cuda_device_count();  // assign to device
-    device           = GetEnv("TOMOPY_DEVICE_NUM", device);
+    device           = GetEnv("TOMOPY_DEVICE_NUM", device) % cuda_device_count();
 
     TIMEMORY_AUTO_TIMER("");
 
--- a/source/libtomo/accel/gpu/sirt.cu
+++ b/source/libtomo/accel/gpu/sirt.cu
@@ -120,6 +120,7 @@ sirt_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
 
     // synchronize the stream (do this frequently to avoid backlog)
     stream_sync(stream);
+    CUDA_FAST_CHECK_LAST_ERROR();
 
     // reset destination arrays (NECESSARY! or will cause NaNs)
     // only do once bc for same theta, same pixels get overwritten
@@ -152,7 +153,7 @@ sirt_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
         CUDA_CHECK_LAST_STREAM_ERROR(stream);
 
         // synchronize the stream (do this frequently to avoid backlog)
-        stream_sync(stream);
+        // stream_sync(stream);
     }
 }
 
@@ -172,7 +173,7 @@ sirt_cuda(const float* cpu_data, int dy, int dt, int dx, const float* center,
     // compute some properties (expected python threads, max threads, device assignment)
     int pythread_num = ntid++;
     int device       = pythread_num % cuda_device_count();  // assign to device
-    device           = GetEnv("TOMOPY_DEVICE_NUM", device);
+    device           = GetEnv("TOMOPY_DEVICE_NUM", device) % cuda_device_count();
 
     TIMEMORY_AUTO_TIMER("");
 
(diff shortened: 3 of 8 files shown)


jrmadsen:
@carterbox Do you use something like `black` for formatting? Or do I have to manually format?

carterbox:
Hi @jrmadsen, good to hear from you!

I use `yapf`, but feel free to use `black`. The pep8speaks bot should only be complaining about lines that have been touched by this PR.

carterbox (changes requested):
Looks like your changes are causing all of the reconstruction methods to return all `1e-6`s. I re-ran the CI on the [upstream branch](https://dev.azure.com/tomopy/tomopy/_build/results?buildId=762&view=results) just to confirm this isn't related to our dependencies.

jrmadsen:
@carterbox Good to talk to you too. Are you staff yet at ANL? Working on anything else beside tomopy?

Those failures might have been related to the fact that I wasn't using `% cuda_device_count()` so it was probably assigning to GPU's that didn't exist. The thread-pool executor is that same so not sure why it would affect other tests, unless they are somehow entering the block. I noticed something weird where `kwargs["accelerated"]` is actually `array(bool)`? Maybe they are entering the wrong block bc of that. E.g.:

```python
    print(f"kwargs: {kwargs}")
```

produces:

```console
kwargs: {'accelerated': array(True), 'num_iter': array(10), 'num_gridx': 2048, 'num_gridy': 2048, 'pool_size': 0, 'interpolation': 'NN', 'device': 'gpu', 'grid_size': array([0, 0, 0], dtype=int32), 'block_size': array([32, 32,  1], dtype=int32)}
```

So: `kwargs["accelerated"] is True` evaluates to false b/c of that.

Also, ignore the previous numbers. I eliminated at lot of the synchronizations and reduced the runtime by approximately 3.3x. Here are the updated numbers:

|   | Algorithm | Filename | Cores | Iterations | # Slices | Dimensions | Time to Solution (sec) |
|--| ---------- | --------- | ----- | ---------- | ------- | ------------ | ---------------------- | 
| Edison | sirt | tomo_0001 | 48 | 100 | 24 | 2048 x 2048 | 28069 |
| Edison | mlem | | | | |  | 24480 |
| Perlmutter | sirt (NN) | | 32 | | |  | 177 |
| Perlmutter | mlem (NN) | | | | |  | 178 |

source/tomopy/recon/algorithm.py
```diff
@@ -392,10 +415,29 @@ def _dist_recon(tomo, center, recon, algorithm, args, kwargs, ncore, nchunk):
...
     else:
-        # execute recon on ncore threads
-        with cf.ThreadPoolExecutor(ncore) as e:
-            for slc in use_slcs:
-                e.submit(algorithm, tomo[slc], center[slc], recon[slc], *args, **kwargs)
+        # execute recon on ncore processes. Accelerated methods have internal thread-pool
+        # and NVIDIA NPP library does not support simulatenously leveraging multiple devices
+        # within the same process
+        if "accelerated" in kwargs and (kwargs["accelerated"] is True
+                                        or kwargs["accelerated"]):
```
carterbox, line 422:
```suggestion
        if "accelerated" in kwargs and kwargs["accelerated"]:
```
With Python there is a subtle difference between identity (`is`) and equality (`==`). For example:

```
True is True
not (True is np.array(True))
True == np.array(True)
True == 1
not (True is 1)
```
The existence of `array(True)` means that somewhere in the stack, the inputs are being converted to NumPy arrays.

source/tomopy/recon/algorithm.py
```diff
@@ -392,10 +415,28 @@ def _dist_recon(tomo, center, recon, algorithm, args, kwargs, ncore, nchunk):
...
+            for f, slc in zip(futures, use_slcs):
+                if f.exception() is not None:
+                    raise f.exception()
+                recon[slc] = f.result()
+        else:
+            # execute recon on ncore threads
+            with cf.ThreadPoolExecutor(ncore) as e:
+                for slc in use_slcs:
+                    e.submit(algorithm, tomo[slc], tomo[slc], center[slc],
+                             recon[slc], *args, **kwargs)
```
carterbox, line 439:
```suggestion
                    e.submit(algorithm, tomo[slc], center[slc],
                             recon[slc], *args, **kwargs)
```
This argument is should not be duplicated.

jrmadsen: Haha, yep, that'll do it.

source/libtomo/accel/gpu/mlem.cu
```diff
@@ -156,7 +157,7 @@ mlem_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
         CUDA_CHECK_LAST_STREAM_ERROR(stream);
 
         // synchronize the stream (do this frequently to avoid backlog)
-        stream_sync(stream);
+        // stream_sync(stream);
```
carterbox, line 160: Is `stream_sync` supposed to be disabled here?

source/libtomo/accel/gpu/utils.cu
```diff
@@ -176,8 +183,32 @@ cuda_rotate_kernel(int32_t* dst, const int32_t* src, const float theta_rad,
...
+        else
+        {
+            --_npp_stream_sync;
+        }
+    }
+
+    if(nppGetStream() != stream)
+        throw std::runtime_error("Error! Wrong stream!");
+    // stream_sync(stream);
+    // nppSetStream(stream);
```
carterbox, line 211:
```suggestion
```

source/libtomo/accel/gpu/sirt.cu
```diff
@@ -152,7 +153,7 @@ sirt_gpu_compute_projection(data_array_t& gpu_data, int p, int dy, int dt, int d
         CUDA_CHECK_LAST_STREAM_ERROR(stream);
 
         // synchronize the stream (do this frequently to avoid backlog)
-        stream_sync(stream);
+        // stream_sync(stream);
```
carterbox, line 156: Also this one?

source/libtomo/accel/gpu/common.cu
```diff
@@ -284,8 +284,15 @@ cuda_device_query()
     else
         printf("Detected %d CUDA capable devices\n", deviceCount);
 
+    int specific_device = GetEnv("TOMOPY_DEVICE_NUM", -1);
```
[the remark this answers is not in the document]
jrmadsen: No, this env variable is effectively the way that a GPU number gets assigned in multi-gpu node: each sub-process in the pool gets passed an index and it sets `"TOMOPY_DEVICE_NUM"` via that index. But I put this here because the migration from threads to processes resulted in every process printing out every single GPU, i.e. when I ran at NERSC, I ended up with 64 printouts of 8 GPUs when 8 processes were started.

source/libtomo/accel/gpu/mlem.cu
jrmadsen: Yes. Based on that comment and my recollection, that stream sync was a by-product of a bottleneck of too many kernel launches getting queued up in the stream. No longer the case since the introduction of processes reduced the contention for submitting kernel launches.

source/libtomo/accel/gpu/utils.cu
jrmadsen: This was actually a bottleneck. The revised numbers that I noted (which reduced the recon time from ~570 seconds to 177 seconds) were solely a by-product on the scheme I implemented here with the `_npp_stream_sync`. This new scheme switches the stream in NPP less often and ensures that another thread does not swap out the NPP stream in the middle of this algorithm, which is why I was able to remove so many synchronizations.

source/libtomo/accel/gpu/sirt.cu
jrmadsen: Same response as mlem and the aformentioned `__npp_stream_sync` impl.

source/libtomo/accel/gpu/common.cu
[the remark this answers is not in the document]
jrmadsen: Here is the relevant impl in algorithm.py:

```python
def _run_accel_algorithm(idx, _func, tomo, center, recon, *_args, **_kwargs):
    # set the device to the specified index within the new process
    os.environ["TOMOPY_DEVICE_NUM"] = "{}".format(idx)
    # execute the accelerated algorithm
    _func(tomo, center, recon, *_args, **_kwargs)
    # recon is the only important modified data
    return recon
```

As you can see, even if the user were to specify this env variable, it would be ignored.

carterbox (approved):
I'd like to see the dead code removed instead of just commented out, but otherwise, looks good to me! Thanks, @jrmadsen!

P.S. I am staff at ANL now! :smile: My primary responsibility is to develop algorithms and implementations for ptychotomographic imaging. The code for that project is https://github.com/tomography/tike.

PR #531, opened 2021-03-19
BLD: Target all CUDA architectures >= 3.0 by default

carterbox:
The purpose of this PR is to change the default behavior of the build system to target all CUDA architectures from 3.0 to 8.6 inclusive instead of just 3.5. The reason for this change is that we distribute a precompiled binary on conda-forge, so most users aren't compiling for themselves. The consequence is the the binary will become larger, but users with arch=30 will be supported and most should get a performance bump?

My approach is to use the `-gencode=arch=compute_${ARCH},code=sm_${ARCH}` flags for NVCC, and enumerate a list of the compatible architectures for CUDA 11 and 9,10. Online documentation for CUDA <= 8 is no longer available, so I didn't consider that case.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -23,6 +23,7 @@ endif()
 cmake_policy(SET CMP0048 NEW)
 cmake_policy(SET CMP0042 NEW)
 cmake_policy(SET CMP0053 NEW)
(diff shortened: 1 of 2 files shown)


jrmadsen:
Looks good to me

jrmadsen (approved):
You could though use the CMake cuda archs for this but I understand not wanting to require such a new cmake version

carterbox:
Thanks, @jrmadsen!

PR #533, opened 2021-03-20
Fix for issue #527: print statements in source/tomopy/recon/algorithm.py

### LINKED ISSUE #527: change verbosity?
**Describe the problem**
I would like to stop displaying things like:

Reconstructing 1 slice groups with 1 master threads...

over and over. There must be an option but I can't find it.

aps-7bm:
Converts print statements about reconstruction initialization to logging and warning. Closes #527.

Also fixes warnings on build due to np.int statements and the use of imp rather than importlib.

Edits to envs/linux-38.yml to allow for builds from source in python 3.8.


aps-7bm:
Added initialization to zero for the reconstruction array for gridrec, as mentioned in the discussion of issue #494.  I also changed the logic of the remove_outliers function so that it performs the initial median filtering in the dtype of the input array (this would usually be int), rather than casting to float32 first.

carterbox (changes requested):
Hi @aps-7bm, thanks for the patch. In addition to the issue below, please revert your changes to `envs/` because I have already updated the environments upstream.

aps-7bm:
I made your requested changes to envs/linux-38.yml (I merged from upstream) and changed logger.warning to warnings.warn in source/tomopy/recon/algorithm.py.

On the subject of the initialization of the recon array, I looked through the C code for gridrec, and the initial value should not matter.  I changed the logic in algorithm.py a bit to use np.empty rather than np.zeros.  I did not see an improvement in speed, though much of the time in my code is used in other tasks (outlier removal, beam hardening corrections), so there might be a slight improvement that I could not see.

carterbox (approved):
Thanks, @aps-7bm! Please let me know if you are planning any other patches; otherwise, I'll make a release with these changes this week.

aps-7bm:
No, I think I'm done for now.  Thanks!

PR #546, opened 2021-06-11
Paganin correction

Arhatari:
Paganin's filter is corrected according to the equation in the original paper. 
Frequency domain is normalized to the value of 2*pi instead of original value of 0.5.
When padding is set to False, the result was giving constant value on the image. Therefore the code is changed accordingly.


Arhatari:
closing this request and opening a new one with additional changes

carterbox:
@Arhatari For the future, adding or subtracting commits doesn't require a new pull request (PR). If a PR is open and you change the branch for the pull request, the changes will automatically appear here.

PR #547, opened 2021-06-11
Paganin filter correction

Arhatari:
Paganin's filter is corrected according to the equation in the original paper.
Frequency domain is normalized to the value of 2*pi instead of original value of 0.5.
When padding is set to False, the result was giving constant value on the image. Therefore the code is changed accordingly.
Minus logarithm is added according to the Paganin's filter equation. (The place of the minus logarithm could also be in the filter itself).

### DIFF
--- a/source/tomopy/prep/phase.py
+++ b/source/tomopy/prep/phase.py
@@ -78,7 +78,7 @@ PLANCK_CONSTANT = 6.58211928e-19  # [keV*s]
 
 def _wavelength(energy):
     return 2 * PI * PLANCK_CONSTANT * SPEED_OF_LIGHT / energy
-
+    
 
 def retrieve_phase(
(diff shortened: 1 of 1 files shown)


dgursoy:
Thanks @Arhatari for this PR! This has been brought up several times before, but never had the chance to do it properly.

However, because there are many users that may have used and become familiar with the parameter settings of the "current" filter, I suggest we name the new function as ```retrieve_phase_paganin``` so that some of the users want to continue using their old function. We may need to update the documentation and say which what each function actually does. What do you say?

Arhatari:
Yes. I agree with your suggestion. This is a great suggestion and will be useful for the users.
Thank you for responding this matter.

Best regards
Benedicta
UNOFFICIAL

PR #548, opened 2021-06-20
Fix link in example

kolibril13:
Here:
https://tomopy.readthedocs.io/en/latest/ipynb/tomopy.html#Reconstruction-with-TomoPy
Is currently a broken link to the tomo data, that is fixed by this pr.

### DIFF
--- a/doc/source/ipynb/tomopy.ipynb
+++ b/doc/source/ipynb/tomopy.ipynb
@@ -74,7 +74,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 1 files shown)


kolibril13:
Just to mention it, there is a very easy option to add readthedocs builds to the GitHub pipeline for pull requests:
https://docs.readthedocs.io/en/stable/pull-requests.html
image

carterbox (approved):
Hi @kolibril13,

I agree that it's better to have an http link instead of a relative path in this case.

Also, thanks for the hint about readthedocs PR validation! Using that features should make it easier to remember to check the docs when we make changes.

Issue #415, opened 2019-06-04
Ring Removal after reconstruction


ankitaeclipse:
Hi,

I am looking to remove ring artefacts from my reconstructed micro-CT images. Is it possible to do this using the remove stripe functionality in tomopy or is it restricted to during reconstruction process?

Thank you for your time.

Kind Regards,
Ankita Singh

carterbox:
Hi @ankitaeclipse,

The post-reconstruction ring removal function we have is here:
https://tomopy.readthedocs.io/en/latest/api/tomopy.misc.corr.html#tomopy.misc.corr.remove_ring

ankitaeclipse:
Hi @carterbox,

Thank you for posting the link. 

1) The function requires some threshold parameters: thresh, thresh_max, thresh_min. How are these parameters decided?
2) What should be the quantization level of the reconstructed image that is given as an input to the function? 

Kind Regards,
Ankita

decarlof:
remove stripe work on the sinogram => is applied before the recontruction

ankitaeclipse:
Hi @decarlof 

Thank you for letting me know. 

I just want to clarify: https://tomopy.readthedocs.io/en/latest/api/tomopy.misc.corr.html#tomopy.misc.corr.remove_ring

refers to ring removal before reconstruction?

carterbox:
The one ring removal method we have is for removing rings post-reconstruction. That is what you initially asked for. Stripe removal methods that @decarlof is suggesting take a sinogram as input. They are used before reconstruction.

1. The rings that you see in your reconstructed image have some brightness value. The thresholds are used to tell the algorithm in what value range to look for the rings, so it doesn't correct things that are not rings.

2. It looks like this function takes floating point valued images, so you can use any bit-depth you want.

decarlof:
no  https://tomopy.readthedocs.io/en/latest/api/tomopy.misc.corr.html#tomopy.misc.corr.remove_ring

works on the reconstructed images

ankitaeclipse:
Thank you @decarlof and @carterbox for your feedback. I will give this a try and get back to you on how I go.

ankitaeclipse:
Hi @carterbox  @decarlof 

I tried the tomopy.misc.corr.remove_ring function. I get the following error:

![Ring Removal Error](https://user-images.githubusercontent.com/17406040/58931215-311dd080-87a2-11e9-96fe-8b2f1753242e.JPG)

carterbox:
[...]

PR #552, opened 2021-07-04
551 fix remove ring theta min

### LINKED ISSUE #551: Crash in remove_ring() with large theta_min
**Describe the problem**

The tomopy.remove_ring() function will crash with a seg fault if a large value of theta_min is given.

**To Reproduce (if applicable)**

The following will crash with a segfault in mean_filter_fast_1D.

```python
import tomopy
import numpy

data = numpy.zeros([1,50, 50])
result = tomopy.remove_ring(data,thresh=7,
                           thresh_max=1,
                           thresh_min=0,
                           rwidth=8,
                           theta_min=158)
```

```
Segmentation fault (core dumped)
```

This is due to an out of bounds access to the image array:
```
0x00007fffe3061086 in mean_filter_fast_1D (filtered_image=0x7fff7433fbc8, image=0x7fff7433fc60, start_row=0, start_col=16, end_row=156, end_col=24, int_mode=0, kernel_rad=158, width=25, height=157) at ../../../source/libtomo/misc/remove_ring.c:419
419	                sum += image[0][row][col];
(gdb) print row
$1 = -1
```

This happens when kernel_rad is larger than height. There are checks to wrap/reflect if row is less than 0 or greater than height, but they only allow wrapping once. A kernel_rad equal to the height is effectively setting the minimum ring size to a full 360 degree circle. However due to the rounding in how kernel_rad is calculated the crash happens with smaller numbers when the image size is small. 

kernel_rad is calulated from the theta_min parameter (minimum angle in degrees (int) to be considered ring artifact) to tomopy.misc.corr.remove_ring(). This gets passed to the angular_min argument in remove_ring in remove_ring.c. Here it is scaled to the size of the polar transformed image,

m_azi = ceil((float) pol_height / 360.0) * angular_min;

For the values above, pol_hight is 50*pi=157,  ceil((float) pol_height / 360.0) = 1, so kernel_rad get set to 158.

**Platform Information:**
 - OS: Ubuntu
 - Python Version 3.8
 - TomoPy Version 1.9 and current master

**Additional context**

Patch to follow

samtygier-stfc:
Fixes #551 by improving the calculation for m_azi, and preventing theta_min angles larger than 360 being passed to remove_ring.

### DIFF
--- a/source/libtomo/misc/remove_ring.c
+++ b/source/libtomo/misc/remove_ring.c
@@ -629,7 +629,7 @@ remove_ring(float* data, float center_x, float center_y, int dx, int dy, int dz,
         polar_image =
             polar_transform(image, center_x, center_y, dx, dy, &pol_width, &pol_height,
                             thresh_max, thresh_min, r_scale, ang_scale, ring_width);
-        m_azi = ceil((float) pol_height / 360.0) * angular_min;
+        m_azi = floor((float) pol_height / 360.0 * angular_min);
         m_rad = 2 * ring_width + 1;
 
         // Call Ring Algorithm
--- a/source/tomopy/misc/corr.py
+++ b/source/tomopy/misc/corr.py
@@ -557,7 +557,7 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
     thresh_min : float, optional
         min value for portion of image to filer
     theta_min : int, optional
-        minimum angle in degrees (int) to be considered ring artifact
+        minimum half angle in degrees (int) to be considered ring artifact
     rwidth : int, optional
         Maximum width of the rings to be filtered in pixels
     int_mode : str, optional
@@ -597,6 +597,9 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
         int_mode = 1
     else:
         raise ValueError("int_mode should be WRAP or REFLECT")
+
+    if not 0 <= theta_min < 180:
+        raise ValueError("theta_min should be in the range [0 - 180)")
 
     args = (center_x, center_y, dx, dy, dz, thresh_max, thresh_min,
             thresh, theta_min, rwidth, int_mode)
--- a/source/tomopy/misc/corr.py
+++ b/source/tomopy/misc/corr.py
@@ -557,7 +557,7 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
     thresh_min : float, optional
         min value for portion of image to filer
     theta_min : int, optional
-        minimum half angle in degrees (int) to be considered ring artifact
+        features larger than twice this angle (degrees) will be considered a ring artifact
     rwidth : int, optional
         Maximum width of the rings to be filtered in pixels
     int_mode : str, optional
--- a/source/tomopy/misc/corr.py
+++ b/source/tomopy/misc/corr.py
@@ -557,7 +557,8 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
     thresh_min : float, optional
         min value for portion of image to filer
     theta_min : int, optional
-        features larger than twice this angle (degrees) will be considered a ring artifact
+        Features larger than twice this angle (degrees) will be considered
+        a ring artifact. Must be less than 180 degrees.
     rwidth : int, optional
         Maximum width of the rings to be filtered in pixels
     int_mode : str, optional


samtygier-stfc:
Questions: Should the limit actually be 0 - 180? The loop for averaging goes to plus and minus the angle

```c
for(int n = -kernel_rad; n < (kernel_rad + 1); n++)
```

What is the right thing to do in the `int_mode == INT_MODE_REFLECT` case. This reflects at `height /2` and so crashes with smaller numbers. The attached changes improve things but safe range is smaller. I am not sure of the use case for this so have not looked into the details.

dyparkinson:
I can only give a hand-waving comment, which is that we use this ring removal once in a while and I've never seen anybody set theta_min to more than 90. So limiting this to 0-180 seems totally fine to me.

carterbox:
As mentioned here: https://github.com/tomopy/tomopy/issues/415#issuecomment-499360659 and in the source code from many commits ago https://github.com/tomopy/tomopy/commit/46e646f77bd9fb7097ab2c70dcf60b988a551dc4 by @lbluque. The original sources may inform the intended behavior.
```
 /*
 This is code for a ring artifact removal algorithm that works in the reconstructed domain.
 It's based on the paper "Comparison of ring artifact correction methods for flat-detector CT" by Prell, et al. It works by:
    -Performing a polar transform on the reconstructed image.
    -Doing a median filter in the radial direction (with a kernel width that depends on the radius).
    -Taking the difference between the filtered and original polar image.
    -Performing a mean filter in the azimuthal direction (again with a kernel width that depends on the radius).
    -Doing an inverse polar transformation (giving an image of just the rings).
    -Subtracting the ring image from the original image.
 Original author: Justin Blair
 */
```

samtygier-stfc:
@carterbox it was the reflect mode specifically that I don't know the use case for. I don't see it in the paper.

The patch fixes the wrap mode and shouldn't make the reflect mode worse.

carterbox:
OK. Thanks for the pull! I'll have time for a proper review next week.

source/tomopy/misc/corr.py
```diff
@@ -598,6 +598,9 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
     else:
         raise ValueError("int_mode should be WRAP or REFLECT")
 
+    if not 0 <= theta_min <= 360:
+        raise ValueError("theta_min should be in the range 0 - 360")
+
```
carterbox, line 603: My understanding is that `theta_min` in the python API is the same a `angular_min` in the C API. Therefore, `theta_min` determines `m_azi`. I agree that it is sensible to check that `theta_min` must be <= zero and < 360 because a ring artifact cannot be longer than 360 degrees. However, is `theta_min` = 360 valid?

carterbox, line 603: My understanding is that `theta_min` in the python API is the same a `angular_min` in the C API. Therefore, `theta_min` determines `m_azi`. I agree that it is sensible to check that `theta_min` must be <= zero and < 360 because a ring artifact cannot be longer than 360 degrees. However, is `theta_min` = 360 valid?

source/libtomo/misc/remove_ring.c
```diff
@@ -629,7 +629,7 @@ remove_ring(float* data, float center_x, float center_y, int dx, int dy, int dz,
         polar_image =
             polar_transform(image, center_x, center_y, dx, dy, &pol_width, &pol_height,
                             thresh_max, thresh_min, r_scale, ang_scale, ring_width);
-        m_azi = ceil((float) pol_height / 360.0) * angular_min;
+        m_azi = floor((float) pol_height / 360.0 * angular_min);
```
carterbox, line 632: Could you explain more about why changing from ceil to floor relates to L273 and where the out of bounds error actually occurs?

```suggestion
        m_azi = ceil((float) pol_height / 360.0 * angular_min);
        // angular_min is float instead of int
```
Would switching the type of `angular_min` to allow `float` inputs also work? This way would be backward compatible.

source/tomopy/misc/corr.py
carterbox: Since `m_azi` is actually the *radius* of the averaging kernel and not the width (the docs should be fixed to clarify). The limit of `theta_min` should actually be strictly less than 180. We can prove this by looking at the code which assumes the following:
```
kernel_rad + kernel_rad + 1 <= height  // pol_height is int height; int kernel_rad is int m_azi
m_azi = ceil((float) pol_height / 360.0) * angular_min;
```
which may be combined to show that `angular_min < 180`.

source/libtomo/misc/remove_ring.c
carterbox: We must switch from ceil to floor because otherwise, the strictly less condition is not met for heights smaller than 179. i.e.
```
ceil((float) height / 360.0 * (2 * 179)) < height
```
is not true for small height.

carterbox (changes requested):
I was also not able to figure out "reflect" mode either; I don't understand why it reflects at height / 2 instead of height.

As discussed in the comments, the ValueError should actually be raised when theta_min is >= 180.

Could you please also edit the docstring for remove_ring, so it clarifies that the `theta_min` refers to the radius (half arclength?) of the rings? Thanks!

source/tomopy/misc/corr.py
carterbox:
```suggestion
    if not 0 <= theta_min < 180:
        raise ValueError("theta_min should be in the range [0, 180)")

```

samtygier-stfc: I agree that 180 is a better limit, especially if in practice much smaller values are generally used.

samtygier-stfc:
> Could you please also edit the docstring for remove_ring, so it clarifies that the theta_min refers to the radius (half arclength?) of the rings? Thanks!

I have changed it to "half angle" because it is still in units of angle. I think radius or length would imply units of pixels.

source/tomopy/misc/corr.py
```diff
@@ -557,7 +557,7 @@ def remove_ring(rec, center_x=None, center_y=None, thresh=300.0,
     thresh_min : float, optional
         min value for portion of image to filer
     theta_min : int, optional
-        minimum angle in degrees (int) to be considered ring artifact
+        minimum half angle in degrees (int) to be considered ring artifact
```
carterbox, line 560: Is "half angle" a common phrase? It is not immediately obvious to me that this means we multiply this parameter by two to get the minimum ring artifact size.
```suggestion
        one half of the minimum angle (degrees) to be considered a ring artifact
```
```suggestion
        features larger than twice this angle (degrees) will be considered a ring artifact
```

samtygier-stfc: Not sure if its common but "half angle" is used sometimes. But I'd be happy with any of these suggestions.

Another option would be to divide the angle by 2 inside the function, so that it is the "minimum angle". But that would change the existing behaviour.

carterbox (approved):
Thanks @samtygier-stfc, for finding this bug and helping to publish a patch! 🚀

PR #558, opened 2021-09-14
BLD: Address some issue with CMAKE build system on Windows

### LINKED ISSUE #556: Problem compiling on win 10
Hi everyone

A question from a tomopy newbie :-)

I am trying to compile tomopy for the first time in order to be able to use the GPU version of the reconstruction algorithm.

First of I would be sure that I understood that correctly and that compilation is really mandatory in order to use GPU acceleration... can anyone please confirm this to me ?

Secondly, is this the right place to ask this kind of questions, or should I use a dedicated forum/chat/discord room .... I am not aware of ?

Concerning compilation problem, it arises during the configuring phase of cmake when executing python setup.py build,  and  I get a lot of errors of this type 

"CMake Error at source/PTL/cmake/Modules/MacroUtilities.cmake:887 (add_library):
  Target "ptl-static" links to target "CUDA::nppc_static" but the target was
  not found.  Perhaps a find_package() call is missing for an IMPORTED
  target, or an ALIAS target is missing?
Call Stack (most recent call first):
  source/PTL/source/CMakeLists.txt:60 (build_library)"

cmake finds  the correct version of the CUDA SDK before giving this error and I have followed the instructions on the developer section in order to compile from sources.

I  am using 
- Miniconda with python 3.8.11
- cmake 3.19.6 (installed by conda)
- visual studio 2017
- cuda sdk 11.4 with (all packages installed including vs integration)

Anyone as an idea of what am I doing wrong ?

Thanks in advance,

Cheers,

Pier

carterbox:
Fixes some clashes between CMake variables from TomoPy and PTL that are fixed by renaming the TomoPy CMake variables.

However, the problem reported by #556 actually stems from linking differences with Windows and Unix. Apparently, the CMake `findCUDAToolkit` module does not import static link targets for NPP on Windows. I'm not sure if the better approach is to use the same non-static link targets for both unix and Windows or (as implemented here), switch to the non-static targets only for Windows. Opinions, @jrmadsen?

Closes #556


cmake/Modules/Packages.cmake
```diff
@@ -101,8 +101,8 @@ if(TOMOPY_USE_TIMEMORY)
     find_package(TiMemory)
```
jrmadsen, line 101: You should probably strip out the C and C++ implementation of timemory. The package has changed a lot and I doubt this `find_package` will even work on a case-sensitive file system as I switched to just `timemory` instead of `TiMemory` a very long time ago.

jrmadsen: And, on that note, the python code has changed a lot too. I know for a fact that whatever is in benchmarking is broken if you download the timemory in conda-forge instead of the timemory in the jrmadsen channel

cmake/Modules/Packages.cmake
```diff
@@ -186,12 +186,19 @@ if(TOMOPY_USE_CUDA)
...
     enable_language(CUDA)
 
-    list(APPEND EXTERNAL_LIBRARIES "CUDA::nppc_static;CUDA::npps_static;CUDA::nppig_static;CUDA::nppisu_static")
-    list(APPEND EXTERNAL_INCLUDE_DIRS ${CUDA_INCLUDE_DIRS}
+    if (WIN32)
+        list(APPEND TOMOPY_EXTERNAL_LIBRARIES "CUDA::nppc;CUDA::npps;CUDA::nppig;CUDA::nppisu")
+    else()
+        list(APPEND TOMOPY_EXTERNAL_LIBRARIES "CUDA::nppc_static;CUDA::npps_static;CUDA::nppig_static;CUDA::nppisu_static")
+    endif(WIN32)
+    list(APPEND TOMOPY_EXTERNAL_INCLUDE_DIRS ${CUDA_INCLUDE_DIRS}
```
jrmadsen, line 201:
```suggestion
    foreach(_NPP_LIB nppc npps nppig nppisu)
        if(TARGET CUDA::${_NPP_LIB}_static)
            list(APPEND TOMOPY_EXTERNAL_LIBRARIES CUDA::${_NPP_LIB}_static)
        elseif(TARGET CUDA::${_NPP_LIB})
            list(APPEND TOMOPY_EXTERNAL_LIBRARIES CUDA::${_NPP_LIB})
        else()
            message(FATAL_ERROR "Missing CUDA NPP target: CUDA::${_NPP_LIB} / CUDA::${_NPP_LIB}_static")
        endif()
    endforeach()
    list(APPEND TOMOPY_EXTERNAL_INCLUDE_DIRS ${CUDA_INCLUDE_DIRS}
```

Using an interface target, if would look like this:

```cmake
# create an interface library
add_library(tomopy-cuda-npp INTERFACE)
# create an alias in the tompy namespace which helps make it clear that you want to link to
# a cmake target named tomopy::cuda-npp, not a potential library tomopy-cuda-npp (i.e. libtomopy-cuda-npp.so) 
add_library(tomopy::cuda-npp ALIAS tomopy-cuda-npp)


if(...)
    ...
    foreach(_NPP_LIB nppc npps nppig nppisu)
        if(TARGET CUDA::${_NPP_LIB}_static)
            target_link_libraries(tomopy-cuda-npp INTERFACE CUDA::${_NPP_LIB}_static)
        elseif(TARGET CUDA::${_NPP_LIB})
            target_link_libraries(tomopy-cuda-npp INTERFACE CUDA::${_NPP_LIB})
        else()
            message(FATAL_ERROR "Missing CUDA NPP target: CUDA::${_NPP_LIB} / CUDA::${_NPP_LIB}_static")
        endif()
    endforeach()
    target_include_directories(tomopy-cuda-npp INTERFACE ${CUDA_INCLUDE_DIRS})
    ...
endif()

# later (even if tomopy::cuda-npp is an empty interface library)
target_link_libraries(libtomopy-accel 
    PUBLIC 
        tomopy::cuda-npp
        ...
    PRIVATE ...)
```

jrmadsen (changes requested):
You can use the `if(TARGET <target-name>)` expression to check whether a cmake target exists

cmake/Modules/Packages.cmake
jrmadsen: Or instead of making an interface library for each external library, you could just do:

```cmake
add_library(tomopy-external INTERFACE)
add_library(tomopy::external ALIAS tomopy-external)

...
            target_link_libraries(tomopy-external INTERFACE CUDA::${_NPP_LIB}_static)
...

target_link_libraries(libtomopy-accel 
    PUBLIC 
        tomopy::external
        ...
    PRIVATE ...)
```

cmake/Modules/Packages.cmake
carterbox: Do you remember what version of timemory is compatible? Can I just add a version specification?

jrmadsen: 3.0.0 was the big rewrite

cmake/Modules/Packages.cmake
carterbox: Thanks for explaining interface libraries. I was having trouble finding an explanation on the web.

cmake/Modules/Packages.cmake
jrmadsen: I'll resolve this since this should probably go in a different PR

jrmadsen (approved):

cmake/Modules/Packages.cmake
jrmadsen: Yeah the cmake docs aren't very clear but once you get the hang of them, they are a phenomenal way to bundle optional extensions for compiler flags, include dirs, compile definitions, etc. into something that can be easily added exactly where you want them.

PR #565, opened 2021-10-13
Fixed mismatched function prototype vs. implementation

jrmadsen:


carterbox:
Good catch, @jrmadsen! Looks like this mismatch has always been there.

PR #566, opened 2021-11-01
Adding a Median Filter for Non-Finite Values

### LINKED ISSUE #564: Median Removal Of Nan/Inf Values
**Is your feature request related to a problem? Please describe.**

I would like to propose a way to remove NAN and INF values from arrays that replaces the values with the median value of those around it. Right now all Tomopy has is a way to replace NAN and INF values with a single static value. In some cases, this can cause intense streaking in the reconstructions. 

**Describe the solution you would like.**
Below is the code which can be used to remove NAN and INF values based upon the median value of values associated inside an X by X kernel around the pixel in question. 
```

  def replace_bad_values(data, kernel_selective, verbose):
   
      if verbose:
          print('\n** Removing Bad Values')  
          
      bad_values = np.concatenate((np.argwhere(np.isnan(data)), np.argwhere(np.isinf(data))))
      og_shape = np.shape(data)
      
      for value in tqdm(bad_values, desc='Removing Bad Values'):
          first = value[0]
          second_v1 = value[1] - kernel_selective
          second_v2 = value[1] + kernel_selective + 1
  
          if second_v1 < 0:
              second_v1 = 0
          if second_v2 > og_shape[1]-1:
              second_v2= og_shape[1]-1
  
          third_v1 = value[2] - kernel_selective
          third_v2 = value[2] + kernel_selective + 1
  
          if third_v1 < 0:
              third_v1 = 0
          if third_v2 > og_shape[2]-1:
              third_v2= og_shape[2]-1
              
          new_data = data[first, second_v1:second_v2, third_v1:third_v2]
          new_data = np.asarray(new_data)
          
          mini_data = np.median(new_data[np.isfinite(new_data)])
          
          data[first, value[1], value[2]] = mini_data
          
      return data


```
**Describe alternatives you've considered.**
One can use the Tomopy's remove_nan or remove_inf functions, but this leads to intense streaking in some reconstructions if the wrong value is chosen to replace NAN and INF with.

WilliamJudge94:
The following code has been added to Tomopy to allow for easy removal of non-finite values inside of a 3D array. The code iterates through each projection, finds non-finite values (nan/inf), and replaces them with the median value of the surrounding pixel values based upon the kernel size set by the user. The median value does not include non-finite value in its calculation.

Closes #564


WilliamJudge94:
I took a look at some of the failing tests and they are all due to "no module named tqdm" import errors. I would prefer to have a loading bar for these iterations. I have found a few examples of reconstructions with contain 4mil non-finite values. Taking upwards of 1-2 min to replace all values. If users are not given a loading bar they may think the program has stalled on them.

prjemian:
Add `tqdm` to the [environments](https://github.com/WilliamJudge94/tomopy/tree/master/envs) (or other project requirements).   It's a good addition that would provide such a progress bar.

prjemian:
Also, on the developer side install and use [`flake8`](https://flake8.pycqa.org/en/latest/) before each commit/push to ensure your contributions are consistent with PEP8.

carterbox:
Since tomopy is a library (not an application), it is out-of-scope to include and UI elements. If you want to provide feedback to the user, you may use the logging module from the standard library. Using the logging module allows end users to customize the verbosity and format of logging messages without editing tomopy's source code.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,137 @@ def median_filter_cuda(arr, size=3, axis=0):
     return out
 
 
+def find_nonfinite_values_prj(projection, slc_idx):
```
carterbox, line 271:
```suggestion
def _find_nonfinite_values_prj(projection, slc_idx):
```
Functions that are not intended for external use should have a leading underscore.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,137 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+    idx_of_nonfinite_prj = np.nonzero(boolean_of_nonfinite)
+
+    # Creating an identical shaped array for the z position in the 3D array the prj comes from
+    idx_of_nonfinite_data = np.asarray(
+        [slc_idx] * len(idx_of_nonfinite_prj[0]))
+
+    return idx_of_nonfinite_data, idx_of_nonfinite_prj[0], idx_of_nonfinite_prj[1]
+
+
+def determine_nonfinite_kernel_idxs(x_idx, y_idx, kernel, shape_x, shape_y):
```
carterbox, line 302:
```suggestion
def _determine_nonfinite_kernel_idxs(x_idx, y_idx, kernel, shape_x, shape_y):
```
Functions that are not intended for external use should have a leading underscore.

WilliamJudge94:
That makes sense. Although would you allow me to 1) add in a "disable" variable for the tqdm UI 2) Along with the tomopy standard logging feature?

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,137 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+        # Determining the lower and upper bounds for kernel
+        x_lower, x_higher, y_lower, y_higher = determine_nonfinite_kernel_idxs(x_idx,
+                                                                               y_idx,
+                                                                               kernel,
+                                                                               shape[1],
+                                                                               shape[2])
+
+        # Extracting kernel data and fining finite median
+        kernel_cropped_data = data[prj_idx, x_lower:x_higher, y_lower:y_higher]
+        kernel_cropped_data = np.asarray(kernel_cropped_data)
```
carterbox, line 392: A slice of data is already a numpy array? Why do we need to ensure that it is an array?
```suggestion
```

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,137 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+    # Forcing values to be integers
+    integer_values_pre = [x_idx_lower, x_idx_higher, y_idx_lower, y_idx_higher]
+    integer_values = [int(i) for i in integer_values_pre]
+
+    return integer_values
+
+
+def median_filter_nonfinite(data, kernel=1):
+    """Apply a selective median filter with size kernel to all nonfinite
+       values in the 3D data array.
```
carterbox, line 351: Please update the description to make it clearer that the filtering is 2D (not 3D) and happens slice-by-slice along the 0th(?) axis.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,137 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+    data : 3D nd.array
+        The 3D array of data with nonfinite values in it.
+
+    kernel : int
+        The size of the kernel to be used for a local median filter. 
+
+    Returns
+    -------
+    The corrected 3D array with all nonfinite values removed based upon the local
+    median value defined by the kernel size.
```
carterbox, line 364: Please check the sobel_filter function below for correct formatting. Our documentation pages are generated automatically from this docstring, so they need to have correct spacing and indendation.

carterbox (changes requested):
Thanks for opening your first PR!

- [x] As mentioned above, it is inappropriate to use tqdm in a library.
- [ ] Please add a test of median_filter_nonfinite() in test/test_tomopy/test_misc/test_corr.py. The test should be a simple example which checks that your function does what it claims to do.
- [x] Add your function to `__all__` at the top of the file
- [x] Add your name to the authors at the top of the file

source/tomopy/misc/corr.py
WilliamJudge94: This has been removed in the latest commit (040be55 - Updating median_filter_nonfinite to not require tqdm)

source/tomopy/misc/corr.py
WilliamJudge94: This has been removed in the latest commit (040be55 - Updating median_filter_nonfinite to not require tqdm)

source/tomopy/misc/corr.py
WilliamJudge94: This has been removed in the latest commit (040be55 - Updating median_filter_nonfinite to not require tqdm)

carterbox:
> would you allow me to add in a "disable" variable for the tqdm UI?

No.
#### The Why
If we included tqdm in  our implementation, then tqdm would become a required dependency or lazy import when the end user could easily wrap this function in tqdm with the following:
```python
for i in tqdm(range(len(data))):
    data[(i,)] = filter_median_nonfinite(data[(i,)])
```
while maintaining full control over the many formatting options that tqdm provides.

#### Alternative
filter_median_nonfinite can accept a callback function which is called once per iteration. This is very similar to what you have implemented in 040be55, but instead of passing a tqdm instance, you would pass a lambda function.
```python
with tqdm(total=len(data)) as pbar:

    def callback():
        pbar.update(1)

    data = filter_median_nonfinite(data, callback=callback)
```
This keeps the tomopy codebase simpler and allows end-users to choose another framework besides tqdm.

WilliamJudge94:
1) Since I have two main loops inside this median_filter function would you like me to have two different callbacks? One for each loop? Or would the solution detailed in 2) be sufficent?

2) Since I would need a way to pass an iterator_length into their callback would it be ok if I changed your callback function into a callback class? Something similar to this:

```
class CallbackClass(self):
    
    def callback():
        # Allow user to define this but have access to the iterator length
        
    def iterator_length():
        # pass in length
        
    def iterator_name():
        # pass in name of loop
```

This allows me to define `callback_loop1 = CallbackClass()`,  `callback_loop2 = CallbackClass()` for each loop defined by the function.

WilliamJudge94:
Assuming these commits pass testing, the following is an example of how to properly run the tqdm progress bar with the median_filter_nonfinite() function. I can add this function implementation to the tomopy docs assuming you approve the current implementation. 


```
from tomopy.misc.corr import median_filter_nonfinite
from time import sleep
import numpy as np

data = np.ones(shape=(1000, 1000, 1000))

for i in range(5000):
    x = np.random.randint(0, 100)
    y = np.random.randint(0, 100)
    z = np.random.randint(0, 100)
    data[z, x, y] = np.inf

with tqdm() as pbar:

    def callback(self):
        pbar.total = self.iterator_len
        pbar.set_description(self.iterator_desc)
        pbar.update(1)

        # Callback function is too fast leading to display errors in jupyter
        sleep(0.003)

    data = median_filter_nonfinite(data, callback=callback)
```

carterbox:
Why do we need two progress bars? Process one projection at a time so there is only one progress bar needed?
```python
for p in projections:
    nonfinite_idx_store = _find_nonfinite_values_prj(p)
    for indices in zip(nonfinite_idx_store.transpose()):
        # apply filter
    pbar.update(1)
```

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,168 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+
+
+def median_filter_nonfinite(data, kernel=1, callback=None):
+    """
+    Apply a selective 2D median filter, slice by slice, for a 3D data array
+    with filter_size=kernel to all nonfinite values in the 3D data array.
+
+    Parameters
+    ----------
+    data : 3D nd.array
```
carterbox, line 356:
```suggestion
    data : ndarray
```

carterbox:
Alternatively you could define the callback as:
```python
def callback(total, description):
        pbar.total = total
        pbar.set_description(description)
        pbar.update(1)
```
No one said the callback had to accept exactly one argument.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,168 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+            iterator_len=data_len_loop1, iterator_desc='Finding Bad Values')
+        callback_loop1.callback = callback
+
+    # Iterating throug each projection to save on RAM
+    for idx, projection in enumerate(data):
+        nonfinite_idx_store = np.concatenate((nonfinite_idx_store,
+                                              _find_nonfinite_values_prj(projection,
+                                                                         int(idx))), axis=1)
+        if callback is not None:
+            callback_loop1.callback(callback_loop1)
```
carterbox, line 395: Instead of checking if the callback is None every iteration, always call the callback. If the user doesn't provide a callback, instantiate a function that does nothing:
```python
# This goes at the top
if callback is None:
    def callback(...):
        pass

for x in y:
    # do stuff
     callback(...)
    # the callback is always called regardless of whether the user provided it
```

carterbox (changes requested):

WilliamJudge94:
I have allowed the median_filter_nonfinite() function to run on a single progress bar instead of two. In addition, I have allowed users to define a callback function that does not depend on a CallbackClass. Below is the proper usage of the callback variable.

```
with tqdm() as pbar:

    def callback(total, description, unit):
        pbar.total = total
        pbar.set_description(description)
        pbar.unit = unit
        pbar.update(1)

    data = median_filter_nonfinite(data, callback=callback)

```

test/test_tomopy/test_misc/test_corr.py
```diff
@@ -66,6 +66,21 @@ def test_gaussian_filter(self):
...
         loop_dim(median_filter, read_file('cube.npy'))
 
+    def test_median_filter_nonfinite(self):
+        data_org = np.ones(shape=(100, 100, 100))
+
+        for i in range(50):
+            x = np.random.randint(0, 100)
+            y = np.random.randint(0, 100)
+            z = np.random.randint(0, 100)
+            data_org[z, x, y] = np.inf
```
carterbox, line 76:
```suggestion
        for i in range(50):
            x = np.random.randint(0, 100)
            y = np.random.randint(0, 100)
            z = np.random.randint(0, 100)
            data_org[z, x, y] = np.inf
            x = np.random.randint(0, 100)
            y = np.random.randint(0, 100)
            z = np.random.randint(0, 100)
            data_org[z, x, y] = -np.inf
            x = np.random.randint(0, 100)
            y = np.random.randint(0, 100)
            z = np.random.randint(0, 100)
            data_org[z, x, y] = np.nan
```
We need to test all types of nonfinite values.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,155 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+    integer_values = [int(i) for i in integer_values_pre]
+
+    return integer_values
+
+
+def median_filter_nonfinite(data, kernel=1, callback=None):
+    """
+    Apply a selective 2D median filter, slice by slice, for a 3D data array
+    with filter_size=kernel to all nonfinite values in the 3D data array.
+
```
carterbox, line 353:
```suggestion
    """
    Remove nonfinite values from a 3D array using an in-place 2D median filter.
    
    The 2D selective median filter is applied along the last two axes of
    the array.

```

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,155 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+    data : ndarray
+        The 3D array of data with nonfinite values in it.
+    kernel : int
+        The size of the kernel to be used for a local median filter. 
+    callback : func
+        A callback function to be run for each loop iteration.
+        Define total, description, and unit variables inside the function.
+        Where total is the total length of the progress bar,
+        description is the description of the progress bar,
+        and unit is the iteration unit of the progress bar.
```
carterbox, line 365:
```suggestion
    callback : func(total, description, unit)
        A function called after every internal loop iteration.
        total is number of loop iterations.
        description is 'Nonfinite median filter'.
        unit is ' prjs'.
```

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,114 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+        x_idx_higher = shape_x
+
+    # Determining y kernel
+    y_idx_lower = y_idx - kernel
+    y_idx_higher = y_idx + kernel + 1
+
+    if y_idx_lower < 0:
+        y_idx_lower = 0
+    if y_idx_higher > shape_y:
+        y_idx_higher = shape_y
```
carterbox, line 308: These 12 lines can be reduce to 4 lines using the built-in `min(a,b)` and `max(a,b)` functions.

source/tomopy/misc/corr.py
```diff
@@ -267,6 +268,114 @@ def median_filter_cuda(arr, size=3, axis=0):
...
+                x_higher,
+                y_lower,
+                y_higher,
+            ) = _determine_nonfinite_kernel_idxs(
+                x_idx,
+                y_idx,
+                size // 2,
+                data.shape[1],
+                data.shape[2],
+            )
```
carterbox, line 363: Once refactored to just 4 lines,  `_determine_nonfinite_kernel_idxs` can be removed and inlined here.

test/test_tomopy/test_misc/test_corr.py
```diff
@@ -66,22 +73,43 @@ def test_gaussian_filter(self):
...
+            data_org[z, x, y] = np.nan
+
+        data_post_corr = median_filter_nonfinite(
+            data_org.copy(),
+            size=5,
+            callback=None,
+        )
+        # All the post filtering values should be 1 because all of the finite
+        # values are 1.
+        assert np.all(data_post_corr == 1.0)
```
carterbox, line 100: This test could be improved. It should test a small array with a known solution (check the other tests in this module for examples) instead of being a random test with an unknown solution.

carterbox (changes requested):

test/test_tomopy/test_misc/test_corr.py
WilliamJudge94: When coming up with a better test I realized this function breaks if the entire kernel contains non-finite values. I am wondering what you think of the following before I build in error handling.

1) checks if each kernel only contains non-finite values
2) if it does then throw a ValueError to the user asking them to increase kernel size

source/tomopy/misc/corr.py
WilliamJudge94: I am unsure how the min(a, b) and max(a, b) functions help. Both those functions return the absolute min and max values for an array. 

If we take `x_idx=10`, `y_idx=20`, and `kernel=1` we want values of `x_idx_lower=9`, `x_idx_higher=11`, `y_idx_lower=19`, and `y_idx_higher=21`. 

Placing these in an array of `x_array=[0, x_idx_lower, x_idx_higher, max_x_shape]`and running `min(x_array)` returns `0` when we want a return of `x_idx_lower`. 

We could do something like `x_min = min(i for i in x_array if i > 0)`, but when ` x_idx_lower<=0` then the `min()`return is `x_min=x_idx_higher`.

carterbox: For each of the 4 range limits your algorithm is equivalent to something like this:
```python
 x_idx_lower = max(0, x_idx - kernel)
```

carterbox: For clarity, `max` is the function from the standard library which compares two scalars. Not `np.max` which finds the maximum along a dimension of an array. In this context, `x_idx, y_idx` are integer scalars (as documented), so using `np.max` would not be appropriate.

test/test_tomopy/test_misc/test_corr.py
carterbox: Good thinking! Agreed. You can check the length of `kernel_cropped_data[np.isfinite(kernel_cropped_data)]` greater than zero then raise a ValueError. I don't think a warning would be appropriate because returning any non-finite values in the array would mean the function doesn't do as claimed.

Remember to add a `Raises` section to the docstring explaining when a ValueError is raised.

carterbox (approved):
Thanks for your work @WilliamJudge94!

PR #567, opened 2021-11-02
DEV: Prototype for replacing ctypes with pybind11

carterbox:
I'm investigating replacing some of the ctypes interfaces with pybind11 and using OpenMP instead of Python threads, but I ran into some compile-time trouble.

It seems that BuildSetting.cmake manually sets the compiler features instead of letting CMake automatically detect them? This was a problem because pybind11 needs some compiler features that are not included in the manually set list. What is the advantage of setting the compiler features manually?

I'm also considering converting prep.c into a cxx file and just adding the pybind wrapper there instead of having a separate compile/source + linking. Thoughts about keeping the pybind wrapping code separate from the implementation code?


jrmadsen:
> It seems that BuildSetting.cmake manually sets the compiler features instead of letting CMake automatically detect them? This was a problem because pybind11 needs some compiler features that are not included in the manually set list. What is the advantage of setting the compiler features manually?

So specifying compiler features is basically requesting the cmake find a compiler which supports these features and ensure that when something is compiled against previously exported target, make sure the compiler supports these features, e.g. if I specify:

```cmake
target_compile_features(mylibrary INTERFACE cxx_std_17)
```

and then, later, another cmake project does:

```cmake
find_package(mypackage)

target_link_libraries(tomopy-library PRIVATE mypackage::mylibrary)
```

and the current compiler does not support C++17, it should fail or if it does support C++17 and the user has `CMAKE_CXX_STANDARD` set to 14, then when the `tomopy-library` target gets compiled, it will compile with C++17 because its dependency `mypackage::mylibrary` specified it needed C++17 via the `target_compile_features`.

Looking at the code, it looks like I misread the docs and implemented this wrong. Since tomopy is an end-product (i.e. no one is building _against_ tomopy), there is no need for any of this. Just remove the `CMAKE_CXX_COMPILE_FEATURES` -- I implemented this incorrectly and is serves no purpose.

PR #570, opened 2021-11-10
normalize accepting 2D data

### LINKED ISSUE #568: single tiff image for dark or reference: problem with normalization
Normalization does not occur correctly when a single tiff is loaded into tomopy using dxchange.reader.read_tiff.

Load in a single tiff image (of size (X,Y)) using dxchange.reader.read_tiff for your dark and reference image.

Upon normalization,
```python
tomopy.prep.normalize.normalize(arr, flat, dark, cutoff=None, ncore=None, out=None)
```
the background is not properly subtracted. This is because tomopy.prep.normalize.normalize takes the mean value on axis 0.
```python
    arr = dtype.as_float32(arr)
    l = np.float32(1e-6)
    flat = np.mean(flat, axis=0, dtype=np.float32)
    dark = np.mean(dark, axis=0, dtype=np.float32)
```

If the length of the imported tiff np.array.shape is 2 (i.e. (number of X pixels, number of Y pixels)), you should not be normalizing on the first axis. This will result in improper background subtraction. 

I fixed this by adding an axis to the np array of the tiff before normalization.
```python
if len(flat.shape) == 2:
    flat = flat[np.newaxis,:,:]
```

swelborn:
Users may have one flat/dark image for their entire dataset.

dxchange.reader.import_tiff will pull these single image tiffs in, but they will only be in 2D format.

The user should be able to normalize this 2D data as they do with 3D data. If the user does not read that 3D data is required from documentation (or misses it), they will not know why their background is not subtracted.

Closes #568


carterbox:
@samwelborn, you don't need to open a new pull request (PR) to make changes. This PR tracks your branch (samwelborn:master); any changes you make to that branch will appear here.

carterbox (changes requested):
NumPy arrays have a parameter called 'ndim'. Please use 'ndim' instead of len(foo.shape). Also, please implement these changes for *all* functions which take flat, dark as parameters so that the TomoPy API remains consistent.

carterbox:
:smile Thanks for continuing to work on this! When you are ready for additional review, click the re-review button on the right side of this pull request. Otherwise, I'll assume you're not done making changes.

samwelborn:
@carterbox is it OK to also add the averaging type here? there is a discrepancy in the averaging type (mean/median between methods). looking back I see I forgot to put an if statement in the nearfield function in the latest commit. will add if you approve.

source/tomopy/prep/normalize.py
```diff
@@ -100,7 +100,8 @@ def minus_log(arr, ncore=None, out=None):
     return out
 
 
-def normalize(arr, flat, dark, cutoff=None, ncore=None, out=None):
+def normalize(arr, flat, dark, averaging='mean',
+              cutoff=None, ncore=None, out=None):
```
carterbox, line 104: New parameters need to go at the end of the parameter list; otherwise it breaks function calls that do not use keywords.

```suggestion
def normalize(arr, flat, dark, cutoff=None,
              ncore=None, out=None, averaging='mean'):
```

source/tomopy/prep/normalize.py
```diff
@@ -109,9 +110,12 @@ def normalize(arr, flat, dark, cutoff=None, ncore=None, out=None):
     arr : ndarray
         3D stack of projections.
     flat : ndarray
-        3D flat field data.
+        2D or 3D flat field data.
     dark : ndarray
-        3D dark field data.
+        2D or 3D dark field data.
+    averaging : str, optional
+        'mean' or 'median', based on how the flat and dark arrays
```
carterbox, line 117:
```suggestion
        'mean' or 'median', how the flat and dark arrays
```

source/tomopy/prep/normalize.py
```diff
@@ -217,7 +229,7 @@ def normalize_bg(tomo, air=1, ncore=None, nchunk=None):
     return arr
 
 
-def normalize_nf(tomo, flats, dark, flat_loc,
+def normalize_nf(tomo, flats, dark, flat_loc, averaging='mean',
                  cutoff=None, ncore=None, out=None):
```
carterbox, line 233:
```suggestion
def normalize_nf(tomo, flats, dark, flat_loc,
                 cutoff=None, ncore=None, out=None, averaging='mean'):
```

source/tomopy/prep/normalize.py
```diff
@@ -231,9 +243,12 @@ def normalize_nf(tomo, flats, dark, flat_loc,
     flats : ndarray
         3D flat field data.
     dark : ndarray
-        3D dark field data.
+        2D or 3D dark field data.
     flat_loc : list of int
         Indices of flat field data within tomography
+    averaging : str, optional
+        'mean' or 'median', based on how the flat and dark arrays
```
carterbox, line 250:
```suggestion
        'mean' or 'median', how the flat and dark arrays
```

source/tomopy/prep/normalize.py
```diff
@@ -127,8 +131,16 @@ def normalize(arr, flat, dark, cutoff=None, ncore=None, out=None):
...
+    if flat.ndim == 2:
+        flat = flat[np.newaxis, :, :]
+    if dark.ndim == 2:
+        dark = dark[np.newaxis, :, :]
+    if averaging == 'mean':
+        flat = np.mean(flat, axis=0, dtype=np.float32)
+        dark = np.mean(dark, axis=0, dtype=np.float32)
+    elif averaging == 'median':
+        flat = np.median(flat, axis=0, dtype=np.float32)
+        dark = np.median(dark, axis=0, dtype=np.float32)
```
carterbox, line 143:
```suggestion
        dark = np.median(dark, axis=0, dtype=np.float32)
    else:
        raise ValueError(f"'averaging' must be 'mean' or 'median' not {averaging}")
```

source/tomopy/prep/normalize.py
```diff
@@ -248,14 +263,18 @@ def normalize_nf(tomo, flats, dark, flat_loc,
...
     if cutoff is not None:
         cutoff = np.float32(cutoff)
     if out is None:
         out = np.empty_like(tomo)
-
-    dark = np.median(dark, axis=0)
+    if averaging == 'mean':
+        dark = np.mean(dark, axis=0, dtype=np.float32)
+    elif averaging == 'median':
+        dark = np.median(dark, axis=0, dtype=np.float32)
```
carterbox, line 277:
```suggestion
        dark = np.median(dark, axis=0, dtype=np.float32)
    else:
        raise ValueError(f"'averaging' must be 'mean' or 'median' not {averaging}")
```

source/tomopy/prep/normalize.py
carterbox: This documentation claims that the 'averaging' parameter controls how *both* the dark and flats are combined, but it actually only controls how the darks are combined. Please reread the documentation and to ensure that it is consistent with what is implemented.

carterbox (changes requested):
> is it OK to also add the averaging type here? there is a discrepancy in the averaging type (mean/median between methods).

Yes, adding this new feature is fine as long as the previous behavior is retained by default.

> looking back I see I forgot to put an if statement in the nearfield function in the latest commit. will add if you approve.

I don't know what you're talking about here. What is "the nearfield function"?

- [ ] Please use the [`.. versionchanged::` directive](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#paragraph-level-markup) at the bottom of each docstring where you changed something and add a summary of the changes. The next version will be 1.12.0.

PR #574, opened 2021-11-22
Cupy integration

samwelborn:
This is really just a draft (it is not really clean), but since you have time this week, @carterbox please feel free to look through it.

In order to get this to work, you will need to look at the add_to_skimage directory and install those modules in your skimage directory, and install cupy. I will work later on my fork of skimage. If this were to be adopted into tomopy, that custom fork would have to be added to the environment rather than the typical skimage install.

You can ignore my changes to the docs and that other module called tomodata. In addition to working on the cupy integration for  the alignment section, I am making a level of abstraction from the base tomopy commands for user friendliness. Goals of that project will be updated in my tomodata fork's readme (at some point, working on many projects at the moment).


carterbox (commented):
Thanks for the preview! Here's my initial feedback.

First, we definitely cannot ship any functionality that requires the end-user to modify/patch the installation of another package. Our users are not skilled enough to understand site-packages and this approach seems fragile. It would break every time scikit-image was updated and would not be automatically installed when TomoPy was updated.

Instead, we can copy the CuPy port of phase_cross_correlation into a module inside TomoPy, and in the long-term, we can submit that module to the CuPy repository for inclusion in cupyx. If ports of scikit-image are within the scope of cupyx, great! If not, then we will continue to host it here specifically for this functionality.

Second, do you have a dataset that we can downsample to include as a development/debugging dataset for projection alignment type problems? There isn't one currently, but I would be very helpful if there was one that we could use.

PR #585, opened 2022-05-26
BUG: distribute_jobs no longer works on Win/MacOS

carterbox:
```
    def dump(obj, file, protocol=None):
        '''Replacement for pickle.dump() using ForkingPickler.'''
>       ForkingPickler(file, protocol).dump(obj)
E       TypeError: cannot pickle 'memoryview' object
```
This error is occuring on the conda-forge feedstock OSX builds. Research tells me that it is caused by trying to pass memory views instead of numpy arrays to the multiprocessing distrbute_jobs function.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -96,7 +96,7 @@ jobs:
     vmImage: 'macOS-latest'
   strategy:
     matrix:
-      Python37:
+      Python38:
         python.version: '38'
     maxParallel: 4
   steps:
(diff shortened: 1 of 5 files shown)


carterbox:
A recent patch to the numpy library (1.22.4), changed the behavior of np.from_buffer(), but only on non-linux platforms? On these platforms, the "base" attribute of a shared array is now memory view instead of the same as thee array before from_buffer is applied.

carterbox:
@michael-sutherland, do you have any opinions about whether we should refactor the TomoPy implementation of distribute_jobs or pursue the patch in NumPy? I don't really understand the reasons/motivations behind either case.

carterbox:
The API change which causes this issue was patched upstream.

PR #591, opened 2022-09-16
BUG: Don't require absolute path for shared libraries

### LINKED ISSUE #590: ModuleNotFoundError: TomoPy links to compiled components which are installed separately and loaded using ctypes.util.find_library(). A required library, tomo-recon, was not found.
I just created tomopy env successfully in "colab" (Using condacolab for that) but condacolab cannot activate it  and when I import tomopy this error comes up. any solutions?

carterbox:
The purpose of this PR is to refactor `c_shared_lib()`, so it doesn't require that ctypes `find_library()` return an absolute path. Apparently on arch linux, ctypes `LoadLibrary` can still sometimes find the library with just the filename if it is on the path.

An exception is now raised only if `LoadLibrary` cannot find the library OR `find_library()` returns `None`.

Additional bug fixes include removing some bare exceptions and checking the ThreadPool for exception.

@falkmielke, please try this patch to see if it works for you.

Closes #590

### DIFF
--- a/envs/win-37.yml
+++ b/envs/win-37.yml
@@ -22,5 +22,5 @@ dependencies:
   - mkl
   - mkl-devel
   - mkl_fft
-  - opencv>=3.4
+  - opencv>=3.4,<4.2
   - pytest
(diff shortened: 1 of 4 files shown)


falkmielke:
@carterbox thank you, I've pulled the PR and checked out `separate-libtomo`.
- libraries are found!
- if libraries are not installed/compiled, the correct explanation/error is shown

Thank you!

(tested on linux `5.19.8-arch1-1` with `Python 3.10.7`, no conda)

carterbox:
Great! I'll make a new release once I test on Windows.

PR #593, opened 2022-11-25
wrappers.py: UFO import requests version "0"

farhi:
UFO import: a static version 0 of UFO is required which prevents using UFO reconstruction. I suggest to skip this test in order to be able to use UFO from tomopy (else it is useless).

### DIFF
--- a/source/tomopy/recon/wrappers.py
+++ b/source/tomopy/recon/wrappers.py
@@ -291,7 +291,7 @@ def ufo_dfi(tomo, center, recon, theta, **kwargs):
     Reconstruct object using UFO's Direct Fourier pipeline
     """
     import gi
-    gi.require_version('Ufo', '0.0')
+    # gi.require_version('Ufo', '0.0')
     from gi.repository import Ufo
 
     theta = theta[1] - theta[0]


source/tomopy/recon/wrappers.py
```diff
@@ -291,7 +291,7 @@ def ufo_dfi(tomo, center, recon, theta, **kwargs):
     Reconstruct object using UFO's Direct Fourier pipeline
     """
     import gi
-    gi.require_version('Ufo', '0.0')
+    # gi.require_version('Ufo', '0.0')
```
carterbox, line 294: I think version 0 is the only UFO version available. Maybe you meant to only remove the MINOR version?

```suggestion
    gi.require_version('Ufo', '0')
```

Any comments from @matze? This is legacy code that I have never touched.

matze:
Let's pull in @tfarago. But my first question would be what the actual error message looks like.

tfarago:
> Let's pull in @tfarago. But my first question would be what the actual error message looks like.

@matze, why didn't you use the minor version in the file name (e.g. [meson](https://github.com/ufo-kit/ufo-core/blob/master/ufo/meson.build#L100)) I need to understand this before fixing it...

matze:
Very good question that I lack a reasonable answer for. Perhaps we did/do not adhere to any stability guarantees and wanted to avoid giving that impression? But that's just guess work …

source/tomopy/recon/wrappers.py
tfarago: For me this actually creates an error `ValueError: Namespace Ufo not available for version 0`. As @matze suggested, could you elaborate on the error you get please?

tfarago:
Or you didn't want to have a bunch of typelib files with minor changes. In any case, this should be fixed in the future.

carterbox:
Any further comments from @matze , @tfarago on this? A month is a pretty long time for an open PR, and I don't use UFO, so I'm not qualified to review this PR.

tfarago:
I agree with @farhi, there is no particular reason to require a specific version of UFO. So you can just drop the test. @carterbox, your suggested change from `gi.require_version('Ufo', '0.0')` to `i.require_version('Ufo', '0')` gives an error on my Ubuntu 20: `ValueError: Namespace Ufo not available for version 0`.

carterbox:
Can someone please explain the purpose of using `gi.require_version` to check the version of UFO at *runtime*? Couldn't we just remove these runtime checks and add `ufo~=0.0` to the package metadata? Then users can use `pip check` or maybe pip wouldn't even let them install an incompatible version in the first place?

farhi:
I believe UFO is optional, TomoPy can run without. So in my view, UFO should not impose a new requirement/dependency.

carterbox:
> I believe UFO is optional, TomoPy can run without. So in my view, UFO should not impose a new requirement/dependency.

Pip and Conda both support optional dependency declarations nowadays.

farhi:
And by the way, it starts to be a long discussion for not much in fact. Is it so dreadful to avoid checking for a wrong test ? Is there a risk of some kind ? I bet not.

prjemian:
The question should be: _Is it possible to load the namespace without specifying the version?_  The call to [`gobject_introspection.require_version()`](https://pygobject.readthedocs.io/en/latest/guide/api/api.html?highlight=require_version#gi.require_version)

Is `require_foreign()` a valid substitute, allowing a specific version to be omitted?

farhi:
Hi Pete ! Good to see you around.
I approve your suggestion, the main point of all this is to effectively load the namespace. The installation by itself should have been completed in case the user really wants UFO.

tfarago:
> UFO import: a static version 0 of UFO is required which prevents using UFO reconstruction. 

@farhi, you wrote this in the beginning. What is the actual problem please? How does it _prevent_ using UFO? If you get an error, please describe what happens.

> Can someone please explain the purpose of using gi.require_version to check the version of UFO at runtime? 

That's a requirement of PyGObject, nothing can be done about it. But please note this is just a _warning_, not an error, so everything will work if you omit the check.

> Is there a non-0 version of UFO?

No.

> I believe UFO is optional, TomoPy can run without. So in my view, UFO should not impose a new requirement/dependency.

That makes sense, I just checked and UFO is loaded "on demand" in the respective functions [ufo_fbp](https://github.com/tomopy/tomopy/blob/master/source/tomopy/recon/wrappers.py#L247) and [ufo_dfi](https://github.com/tomopy/tomopy/blob/master/source/tomopy/recon/wrappers.py#L289), so unless people try to use UFO without it being installed they won't get any errors.

> The question should be: Is it possible to load the UFO namespace without specifying the version? 

Yes, just drop the `gi.require_version('Ufo', '0.0')`.

Lastly, please note that (from my perspective) there is nothing wrong with the code as it was before the change in this PR. I stripped the `ufo_fbp` function of all unnecessary details and it runs without an issue:
```python
import gi
gi.require_version('Ufo', '0.0')
from gi.repository import Ufo

g = Ufo.TaskGraph()
pm = Ufo.PluginManager()
sched = Ufo.Scheduler()

input_task = Ufo.InputTask()
output_task = Ufo.OutputTask()
fft = pm.get_task('fft')
ifft = pm.get_task('ifft')
fltr = pm.get_task('filter')
backproject = pm.get_task('backproject')

g.connect_nodes(input_task, fft)
g.connect_nodes(fft, fltr)
g.connect_nodes(fltr, ifft)
g.connect_nodes(ifft, backproject)
g.connect_nodes(backproject, output_task)
```

farhi:
OK, I've gone through a number of tests. First, let me state that I'm running a Debian "stable" (Bullseye 11) which is about similar to a Ubuntu 20.04 system.
Then, UFO is installed providing the official packages (apt search).

- gir1.2-ufo-1.0/stable,now 0.16.0.52.gbd831ab-1+b1 amd64 [installed]
- libufo-bin/stable,now 0.16.0.52.gbd831ab-1+b1 amd64 [installed]
- libufo-data/stable,now 0.16.0.52.gbd831ab-1 all [installed]
- libufo-dev/stable,now 0.16.0.52.gbd831ab-1+b1 amd64 [installed]
- libufo1/stable,now 0.16.0.52.gbd831ab-1+b1 amd64 [installed]
- python3-ufolib2/stable,now 0.11.1+dfsg1-1 all [installed,automatic]
- ufo-core-doc/stable,now 0.16.0.52.gbd831ab-1 all [installed]
- ufo-filters/stable,now 0.16.0.159.g483ce16-1 amd64 [installed]
- ufo-filters-data/stable,now 0.16.0.159.g483ce16-1 all [installed]
- ufo-filters-doc/stable,now 0.16.0.159.g483ce16-1 all [installed]
- python3-gi/stable,now 3.38.0-2 amd64 [installed,automatic

**Testing gi with Ufo**

If I do a simple test
```python
import gi
gi.require_version('Ufo', '0.0')
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
<ipython-input-2-6dbe7ac699e6> in <module>
----> 1 gi.require_version('Ufo', '0.0')

/usr/lib/python3/dist-packages/gi/__init__.py in require_version(namespace, version)
    127 
    128     if version not in available_versions:
--> 129         raise ValueError('Namespace %s not available for version %s' %
    130                          (namespace, version))
    131 

ValueError: Namespace Ufo not available for version 0.0
```

So Ufo is available, but not in version 0.

I can drop the `gi.require_version` :
```
import gi
from gi.repository import Ufo
<ipython-input-2-c58f8ba27dd9>:1: PyGIWarning: Ufo was imported without specifying a version first. Use gi.require_version('Ufo', '1.0') before import to ensure that the right version gets loaded.
  from gi.repository import Ufo
```

The conclusion is that version "1.0" of Ufo is loaded. The error is converted into a warning. The `require_version` line is a confirmed as a blocker. As it is difficult to follow all version numbers of dependencies, it is much easier to drop the line I suggest in this PR.

**Tomopy with Ufo**

I have not managed to get a stable environment that both runs TomoPy and UFO. The best I could do is as follows (and I would appreciate more documentation on your help pages in this respect).

Installation:
```
conda create --name tomopy --channel conda-forge tomopy
conda activate tomopy
conda install -c conda-forge dxchange
conda install -c conda-forge matplotlib
conda install -c conda-forge gobject-introspection
conda install -c conda-forge pygobject
```

then a test:
```python
import dxchange
import tomopy
proj, flat, dark, theta = dxchange.read_aps_32id(fname='/home/experiences/grades/farhie/Downloads/tooth.h5', sino=(0,2))
recon = tomopy.recon(proj, theta, center=0, algorithm=tomopy.ufo_fbp, ncore=1)
```
(BTW there is a missing `tomopy.` in the TomoPy/UFO help page at line `recon = tomopy.recon(proj, thet ...`)

This brings an error
- `ValueError: Namespace Ufo not available`. 

I bet the gobject bindings from the system are not available from conda. Damn.
So I would appreciate a lot to get some detailed installation instructions for a Debian class system, in order to have TomoPy and UFO work.

BTW: my group at Synhcrotron SOLEIL has packaged tomopy for Debian, see https://tracker.debian.org/pkg/tomopy This is not fully functional yet. Some algorithms are still lacking (including compiled C code), as well as the dxchange dependency. And we wish to use UFO...

Thanks, Emmanuel.

matze:
> BTW: my group at Synhcrotron SOLEIL has packaged tomopy for Debian, see https://tracker.debian.org/pkg/tomopy 

Your group (or better Emmanuel Picca and Serge Cohen of SOLEIL) also packaged UFO for Debian. As far as I can tell someone decided to publish the libufo 0.16 as [libufo1](https://packages.debian.org/bullseye/libufo1) whereas every version before was published as [libufo0](https://packages.debian.org/buster/libufo0). That major version number also determines the typelib version as you can see [here](https://packages.debian.org/bullseye/amd64/gir1.2-ufo-1.0/filelist).

tfarago:
> So I would appreciate a lot to get some detailed installation instructions for a Debian class system, in order to have TomoPy and UFO work.

You have two more options (which both come with the added value of having the newest UFO version):

1. Use system package manager to install the dependencies and build UFO yourself by cmake or meson, for Ubuntu 20.04 it's a matter of copy-pasting commands from our [manual](https://ufo-core.readthedocs.io/en/latest/install/linux.html#ubuntu-debian-tested-on-ubuntu-20-04-3-lts);
2. Use [docker](https://ufo-core.readthedocs.io/en/latest/install/docker.html) to get a working UFO installation and install tomopy within that.

carterbox:
Closing this issue since it seems that the problem which opened this issue is actually a problem with the UFO Debian package and not a problem with the TomoPy source code. Thanks to @tfarago, @matze for participating.

@farhi, I have no experience with deb/rpm formats, but if your team wants advice/help on [how to compile tomopy](https://tomopy.readthedocs.io/en/stable/devguide.html#building-tomopy), please contact me via email (dching@anl.gov).

PR #594, opened 2022-11-30
Add 3D median filter from Larix

yousefmoazzam:
As suggested by @dkazanc in previous correspondence regarding developers at DLS potentially collaborating on TomoPy, this PR is testing the waters by attempting to integrate the (CPU) 3D median filter from Larix https://github.com/dkazanc/larix.

The main changes are:
- Add Larix as a dependency
- Add a wrapper function for the 3D median filter in Larix
- Add a unit test for the wrapper function
- Include the newly added median filter in the documentation

It's still a work in progress of course, so please feel free to let us know how else we can add to/improve this!

### DIFF
--- a/doc/source/api/tomopy.misc.corr.rst
+++ b/doc/source/api/tomopy.misc.corr.rst
@@ -16,6 +16,7 @@
       median_filter
       median_filter_cuda
       median_filter_nonfinite
+      median_filter_3d
       sobel_filter
       remove_nan
       remove_neg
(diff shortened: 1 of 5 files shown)


carterbox (commented):
Thank you for your interest in improving TomoPy. However, I will not accept this PR for the following reasons:

- `median_filter_3d` is a trivial wrapping function which probably could have been accomplished in one line: `from larix.methods.misc import MEDIAN_FILT as median_filter_3d` unless larix doesn't provide a docstring or has different default values?
- I'm not going to add a required dependency for all TomoPy users for one new function. Especially since `larix` is not available on the conda-forge channel.

PR #596, opened 2023-01-04
Adding median3d and dezinger3d filters into tomopy.misc.corr

dkazanc:
in this commit: 
- Median3d filter implemented in C language, accessible with `from tomopy.misc.corr import median_filter3d` for float32 and uint16 data types
- Dezinger3d filter implemented in C language, accessible with `from tomopy.misc.corr import remove_outlier3d` for float32 and uint16 data types
- unit tests for median and dezinger3d filters

The C functions are implemented using OpenMP API and a special care taken with TomoPy _ncore_ multithreading option. C modules are also accepting  _ncore_ parameter and schedule computing resources accordingly. By default all cores are enabled.


test/test_tomopy/test_misc/test_corr.py
```diff
@@ -109,6 +111,33 @@ def test_median_filter_nonfinite(self):
...
+        A = np.float32(np.zeros((30,40,20)))
+        A[:] = 0.1
+        A[10:20,20:30,:] = 0.5
+        np.random.seed(1)
+        A_noise = A + np.random.normal(loc=0,
+                          scale=0.03 * A,
+                          size=np.shape(A))
+        A_filtered = median_filter3d(np.float32(A_noise))
+        max_filtered = 0.513833
+        assert_allclose(np.max(A_filtered), max_filtered, rtol=1e-06)
```
carterbox, line 126: How do you know that the maximum value of the filtered array is 0.513833 if the noise is random?

Can we compare with a human verifiable result instead. For example:

```
A = np.arange(4*5*6).reshape(4,5,6)
A1 = median_filter3d(A, 1)
np.testing.assert_equal(A, A1)
A3 = median_filter3d(A, 3)
np.testing.assert_equal(truth, A3)

```

source/tomopy/misc/corr.py
```diff
@@ -346,6 +348,88 @@ def callback(total, description, unit):
...
+    """
+    input_type = arr.dtype
+    if (input_type != 'float32') and (input_type != 'uint16'):
+        arr = dtype.as_float32(arr) # silently convert to float32 data type
+    out = np.empty_like(arr)
+
+    if np.ndim(arr) == 3:
+        dz, dy, dx = arr.shape
+    else:
+        raise ValueError("The input array must be a 3D array")
```
carterbox, line 421: 0 and 1 are valid dimension lengths for numpy. What happens if any of these dimensions are <= 1?

carterbox (changes requested):
Thanks for your contribution! I look forward to seeing these functions merged.

source/libtomo/misc/CMakeLists.txt
```diff
@@ -1,7 +1,23 @@
...
+		   ${HEADERS})
+
+find_package(OpenMP REQUIRED)
+if (OPENMP_FOUND)
+    set (CMAKE_C_FLAGS "${CMAKE_C_FLAGS} ${OpenMP_C_FLAGS}")
+    set (CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} ${OpenMP_CXX_FLAGS}")
+    set (CMAKE_EXE_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_EXE_LINKER_FLAGS} ${OpenMP_CXX_FLAGS}")
+    set (CMAKE_SHARED_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_SHARED_LINKER_FLAGS} ${OpenMP_CXX_FLAGS}")
+    set (CMAKE_STATIC_LINKER_FLAGS "${CMAKE_EXE_LINKER_FLAGS} ${OpenMP_STATIC_LINKER_FLAGS} ${OpenMP_CXX_FLAGS}")
+endif()
```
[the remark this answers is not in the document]
carterbox: Actually, we might try this:

```cmake
find_package(OpenMP REQURIED)
if(OpenMP_C_FOUND)
    target_link_libraries(MyTarget PUBLIC OpenMP::OpenMP_C)
endif()
```

source/tomopy/misc/corr.py
```diff
@@ -346,6 +348,88 @@ def callback(total, description, unit):
...
+def median_filter3d(arr, kernel_half_size=1, ncore=None):
+    """
+    Apply 3D median filter to 3D array (C-implementation).
+
+    Parameters
+    ----------
+    arr : ndarray
+        Input 3D array.
+    kernel_half_size : int, optional
+        The half size of the filter's kernel, i.e. 1 results in the full kernel size of 3 x 3 x 3.
```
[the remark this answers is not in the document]
dkazanc: resolved in [77b4340](https://github.com/tomopy/tomopy/pull/596/commits/77b4340d3226f88cd007e08b93b316562c2b1bcf)

source/tomopy/misc/corr.py
```diff
@@ -346,6 +348,88 @@ def callback(total, description, unit):
...
+    ncore : int, optional
+        Number of cores that will be assigned to jobs.
+
+    Returns
+    -------
+    ndarray
+        Median filtered 3D array.
```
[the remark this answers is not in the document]
dkazanc: resolved in [77b4340](https://github.com/tomopy/tomopy/pull/596/commits/77b4340d3226f88cd007e08b93b316562c2b1bcf)

source/tomopy/misc/corr.py
```diff
@@ -346,6 +348,88 @@ def callback(total, description, unit):
...
+    dif : float
+        Expected difference value between outlier value and
+        the median value of the array.
+    ncore : int, optional
+        Number of cores that will be assigned to jobs.
+
+    Returns
+    -------
+    ndarray
+        Dezingered 3D array.
```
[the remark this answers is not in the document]
dkazanc: resolved in [77b4340](https://github.com/tomopy/tomopy/pull/596/commits/77b4340d3226f88cd007e08b93b316562c2b1bcf)

source/tomopy/misc/corr.py
dkazanc: Indeed a good point, I've checked what happens with the function when the lengths are 1, and 0. It doesn't break in both instances and I think we get a 2D filtration in this case (certainly with the length 1). Do you reckon we need to check if we've got a zero length and get a warning?

carterbox:
```c
    for(k = 0; k < dimZ; k++)
    {
        for(j = 0; j < dimY; j++)
        {
            for(i = 0; i < dimX; i++)
            {
                index = ((dimX * dimY) * k + j * dimX + i);
                medfilt3D_float(Input, Output, radius, sizefilter_total, mu_threshold, i,
                                j, k, index, (long) (dimX), (long) (dimY), (long) (dimZ));
            }
        }
    }
```
I think when any of `dimZ`, `dimY`, `dimX` are zero, this loop exits before any filtering is done, so I think raising an Exception or UserWarning is needed. I agree that for singleton dimensions the behavior probably devolves to the 2D case.

carterbox (changes requested):
compare-median

```python
import numpy as np
import timeit
import matplotlib.pyplot as plt

tomopy8_times = []
tomopy3_times = []
tomopy1_times = []
scipy_times = []
widths = np.unique(np.logspace(start=0, stop=6, base=2).astype(int))

for width in widths:

    times = timeit.repeat(
        number=1,
        stmt='tomopy.median_filter3d(x, size=7, ncore=1)',
        setup=f'''
import numpy as np
x = np.random.rand({width}, {width}, {width})
import tomopy
''',
    )
    tomopy1_times.append((np.mean(times), np.std(times)))

    times = timeit.repeat(
        number=1,
        stmt='tomopy.median_filter3d(x, size=7, ncore=3)',
        setup=f'''
import numpy as np
x = np.random.rand({width}, {width}, {width})
import tomopy
''',
    )
    tomopy3_times.append((np.mean(times), np.std(times)))

    times = timeit.repeat(
        number=1,
        stmt='tomopy.median_filter3d(x, size=7, ncore=8)',
        setup=f'''
import numpy as np
x = np.random.rand({width}, {width}, {width})
import tomopy
''',
    )
    tomopy8_times.append((np.mean(times), np.std(times)))

    times = timeit.repeat(
        number=1,
        stmt='scipy.ndimage.median_filter(x, size=7)',
        setup=f'''
import numpy as np
x = np.random.rand({width}, {width}, {width})
import scipy
    ''',
    )
    scipy_times.append((np.mean(times), np.std(times)))

plt.figure()
y, yerr = zip(*tomopy1_times)
plt.errorbar(
    x=widths,
    y=y,
    yerr=yerr,
)
y, yerr = zip(*tomopy3_times)
plt.errorbar(
    x=widths,
    y=y,
    yerr=yerr,
)
y, yerr = zip(*tomopy8_times)
plt.errorbar(
    x=widths,
    y=y,
    yerr=yerr,
)
y, yerr = zip(*scipy_times)
plt.errorbar(
    x=widths,
    y=y,
    yerr=yerr,
)
plt.legend(
    ['tomopy (ncore=1)', 'tomopy (ncore=3)', 'tomopy (ncore=8)', 'scipy'])
plt.title('Wall time of median_filter for cube of random float32')
plt.xlabel('Cube edge length [pixels]')
plt.ylabel('Wall time [s]')
plt.savefig('compare-median.svg')
```

test/test_tomopy/test_misc/test_corr.py
dkazanc: many thanks, it is clearer now! I will go and do the checking as we might need a specific symmetric boundary condition in _scipy.ndimage.median_filter_ to fit the implementation.

source/tomopy/misc/corr.py
dkazanc: Thanks a lot for the benchmark, looks promising! I removed the 2d implementation inside the code but it can be added very easily. I guess `tomopy.misc.corr.median_filter` could be adapted to it elementary. 

I was wondering, would it make sense making it totally generic then? If `axis` parameter is undefined (e.g. None), then the 3d filter is triggered, when the axis is defined then we have a 2d filter. Or you would still prefer to have a separate instance of median_filter3d? cheers

carterbox: Yes, if we can show performance improvement in both 2D and 3D cases without breaking the existing Python API, then I think adding 3D filtering to `tomopy.median_filter()` instead of adding a new function would be preferred. I think users with existing workflows would appreciate the work-free speedups to their 2D workflows. 😄

source/libtomo/misc/CMakeLists.txt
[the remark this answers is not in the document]
dkazanc: hi, just been trying to benchmark some things here and noticed that your suggestion above doesn't seem to enable multithreading unfortunately (cmake ver.3.25). Reverted to the original version brought it back again. It does work for you, however? What could be the problem? thanks

test/test_tomopy/test_misc/test_corr.py
dkazanc: done in [51566bb](https://github.com/tomopy/tomopy/pull/596/commits/51566bb7091d8cbc3fb9b1f0b524a357c3063ebb)

source/tomopy/misc/corr.py
dkazanc: I've added the 2D version of the filter and it can be called from the `median_filter3d` function with a new parameter `axis` (similar to median_filter but by default  equals to `None`). I've done some basic benchmarking against the TomoPy's multithreading version and it was about two times slower, so I think having `median_filter3d` separately makes more sense. I think I know the reason why it is slower, as I perform slicing in Python and then copy slices to the C code which does 2D filtering. Perform slicing in C (I guess by modifying the filters kernel orientation) should be faster. Could be a task for another day...

source/libtomo/misc/CMakeLists.txt
```diff
@@ -1,7 +1,22 @@
...
+  remove_ring.c
+  median_filt3d.c
+  utils.c
+  utils.h
+  ${HEADERS})
+
+find_package(OpenMP REQUIRED)
+if(OpenMP_C_FOUND)
+  target_link_libraries(tomo-misc PUBLIC OpenMP::OpenMP_C)
+endif()
```
carterbox, line 19:
```suggestion
find_package(OpenMP REQUIRED COMPONENTS C)
target_link_libraries(tomo-misc PRIVATE OpenMP::OpenMP_C)
```

source/libtomo/misc/CMakeLists.txt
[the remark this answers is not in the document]
carterbox: It should work for [any CMake >= 3.9 for Linux](https://cliutils.gitlab.io/modern-cmake/chapters/packages/OpenMP.html). The benchmark I shared is evidence that it works for me because otherwise, your implementation wouldn't speed up with increasing number of cores. Can you share your build logs?

I have also added a new suggestion to the code which removes the conditional, OpenMP_C_FOUND. Maybe CMake isn't detecting OpenMP for C, but for other languages then skipping the link?

Do you have build logs you could share?

source/tomopy/misc/corr.py
carterbox: I appreciate your effort to prototype a new 2D median filter, but since it isn't faster than the existing method, please revert those changes. I don't want to confuse users by providing multiple ways to median filtering with the same method. We can revisit speeding up the 2D filtering in a future PR if you have time.

source/libtomo/misc/CMakeLists.txt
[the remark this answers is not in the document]
carterbox: Also, is there a minimum OpenMP version that these features require?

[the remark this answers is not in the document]
dkazanc: thanks, here are my logs bellow, also OpenMP_C gets detected successfully. Those openmp features were already there since version 3.0.
Screenshot from 2023-01-10 10-27-52
 but the multithreading is still disabled in my case when I do:
```
find_package(OpenMP REQUIRED COMPONENTS C)
target_link_libraries(tomo-misc PRIVATE OpenMP::OpenMP_C)
```
What gcc do you use? Don't know if that matters, I've got 7.5.

[CMakeOutput.log](https://github.com/tomopy/tomopy/files/10381568/CMakeOutput.log)
[CMakeError.log](https://github.com/tomopy/tomopy/files/10381570/CMakeError.log)

source/tomopy/misc/corr.py
dkazanc: reverted to 3D filtering only.

carterbox: Thanks!

source/libtomo/misc/CMakeLists.txt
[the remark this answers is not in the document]
carterbox: I think some project-wide flags were overshadowing the local ones. Try again please.

carterbox (approved):
Please do one last check of your code. Note that I did remove some headers from the new C implementations because they seemed to be unused.

carterbox:
Ope, I broke Windows.

source/libtomo/misc/CMakeLists.txt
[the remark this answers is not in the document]
dkazanc: thanks a lot, the recent changed helped.

dkazanc:
Many thanks @carterbox for your help with this PR. It's been a good tutorial for us and will try to keep the suggested formatting and style in our future PRs. We've got few things that we would like to add to TomoPy. 

Regarding the PR, all seems to behave for me and the multithreading performs as expected. I'm getting a good speed-up compare to `scipy.ndimage.median_filter` which does seems to run serially. Happy for it to be merged. Thanks!

PR #601, opened 2023-02-08
BLD: Update PTL to the latest version remove USE_PTL flag

carterbox:
The USE_PTL CMake option has been slightly redundant for a while, since the accel module always includes PTL headers. This refactoring

- moves the PTL submodule to the latest release instead of using an old commit
- removes TOMOPY_USE_PTL as an option and always statically links to PTL
- changes some function calls and adds many PTL:: namespace markers to adapt to the PTL v2 API.

None of the API/ABIs should be affect by this change

Also fixed a missing CXX standard declaration for CXX gridrec

### DIFF
--- a/source/libtomo/accel/common.hh
+++ b/source/libtomo/accel/common.hh
@@ -60,7 +60,7 @@ CreateThreadPool(unique_thread_pool_t& tp, num_threads_t& pool_size)
     {
 #if defined(TOMOPY_USE_PTL)
         // compute some properties (expected python threads, max threads)
-        auto pythreads = GetEnv("TOMOPY_PYTHON_THREADS", HW_CONCURRENCY);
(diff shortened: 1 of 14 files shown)


source/libtomo/accel/common.hh
```diff
@@ -92,19 +93,19 @@ CreateThreadPool(unique_thread_pool_t& tp, num_threads_t& pool_size)
     auto tid = GetThisThreadID();
 
     // initialize the thread-local data information
-    auto& thread_data = ThreadData::GetInstance();
+    auto& thread_data = PTL::ThreadData::GetInstance();
     if(!thread_data)
-        thread_data.reset(new ThreadData(tp.get()));
+        thread_data = new PTL::ThreadData(tp.get());
```
carterbox, line 98: This was the only API transition that I was unsure about. Previously, GetInstance returned a std::unique_ptr, but now it just returns a regular pointer. I think assignment is the correct operation here? @jrmadsen

PR #606, opened 2023-02-22
BLD: Have CMake manage CUDA architectures

carterbox:
Cmake 3.18 adds the CMAKE_CUDA_ARCHITECTURES
variable, which defines the default cuda archs passed to the cuda compiler. Builders can now set this variable in the cache or use CUDAARCHS environment variable if using CMake >=3.20.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,6 @@
 # need CUDA as a first-class language; use newer FindCUDAToolkit
-cmake_minimum_required(VERSION 3.17 FATAL_ERROR)
+# need CMAKE_CUDA_ARCHITECTURES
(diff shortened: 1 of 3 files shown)


carterbox:
@dkazanc, lemme know if this works for you. You will need to set the CUDAARCHS environment variable with the archs you want to target. Though, CMake 3.23 supports keywords like 'all-major', and 'native'

dkazanc:
thanks, will check this out and report.

dkazanc:
@carterbox it does seem to be working for me as with the wrong (obsolete) architecture chosen (35) cmake cannot find the nvcc compiler.  Finds it when I go with 50 arch or higher. cheers.

PR #611, opened 2023-03-21
DOC: Update install directions

carterbox:
Update CUDA variant install directions to latest information


carterbox:
https://github.com/readthedocs/readthedocs.org/issues/10171

PR #616, opened 2023-05-19
BUG: vector reconstruction update fixed

dgursoy:
This is somehow missed because it was never tested. The issue is not visible for ```vector2``` and ```vector3``` functions, that were used in the tutorials.


carterbox:
Is there a simple unit test that can be added which activates this code path?

carterbox (approved):

PR #617, opened 2023-06-08
option `dtype` removed from np.median

userbeats:
possibly wrong call of [`np.median`](https://numpy.org/doc/stable/reference/generated/numpy.median.html) in `normalize.py` corrected removing the option `dtype=np.float32`

### DIFF
--- a/source/tomopy/prep/normalize.py
+++ b/source/tomopy/prep/normalize.py
@@ -150,8 +150,8 @@ def normalize(
         flat = np.mean(flat, axis=0, dtype=np.float32)
         dark = np.mean(dark, axis=0, dtype=np.float32)
     elif averaging == 'median':
-        flat = np.median(flat, axis=0, dtype=np.float32)
-        dark = np.median(dark, axis=0, dtype=np.float32)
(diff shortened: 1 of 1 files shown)


source/tomopy/prep/normalize.py
```diff
@@ -150,8 +150,8 @@ def normalize(
         flat = np.mean(flat, axis=0, dtype=np.float32)
         dark = np.mean(dark, axis=0, dtype=np.float32)
     elif averaging == 'median':
-        flat = np.median(flat, axis=0, dtype=np.float32)
-        dark = np.median(dark, axis=0, dtype=np.float32)
+        flat = np.median(flat, axis=0)
+        dark = np.median(dark, axis=0)
```
carterbox, line 154:
```suggestion
        flat = np.median(flat, axis=0).astype(np.float32)
        dark = np.median(dark, axis=0).astype(np.float32)
```

carterbox: Median does not have a dtype parameter, but let's preserve the intention of the code.

PR #623, opened 2023-09-24
Fix for remove_outlier3d #622

### LINKED ISSUE #622: Access Violation Error on remove_outlier3d
**Describe the problem**
Hello! I have installed tomopy following the instructions on https://tomopy.readthedocs.io/en/latest/install.html and have successfully used it to reconstruct several test datasets in a conda environment, running the angles, normalize, find_center, minus_log, and recon functions without a problem. However, when trying to dezinger the projection data using remove_outlier3d, I receive the following error:

image

Running remove_outlier works successfully but expectedly slow on a 16 GB file. Any help getting remove_outlier3d to remove hot pixels would be greatly appreciated!

**To Reproduce**
An image is read from an hdf5 file, angle space is calculated, and normalization performed. I'm excluding the latter part of the code which does the reconstruction and writes the output, which works without a problem. The error is thrown whether remove_outlier3d is run before or after the normalization step. The data type of proj_data after normalization is float32, and the normalization step works as expected as confirmed by writing the output and checking in FIJI.

```python
import numpy as np
import tomopy
import dxchange as dx

if __name__ == "__main__":
    # ID file
    fname = 'filename.h5'
    
    # Read projections and flat/dark fields
    proj_data = dx.read_hdf5(fname, "exchange/data")
    wf = dx.read_hdf5(fname, "exchange/data_white")
    df = dx.read_hdf5(fname, "exchange/data_dark")
    
    #Determine total rotation angle of scan, scanDeg can also be manually set
    scanDeg = dx.read_hdf5(fname, "process/acquisition/rotation/rotation_step")
    scanDeg = scanDeg * proj_data.shape[0]

    # Set data collection angles as equally spaced in radians
    theta = tomopy.angles(proj_data.shape[0], ang1=0, ang2=scanDeg)

    # Preprocess data using flat/dark fields
    proj_data = tomopy.normalize(proj_data, wf, df)

    # Dezinger *this is where the error is thrown**************************
    proj_data = tomopy.remove_outlier3d(proj_data, dif=0.5, size=3)
    
```

```
Traceback (most recent call last):
  File "C:\Users\yakovlev\Work\Code\Tomopy_Recon_Test\MSU_Maoz_4A_16GB_360deg\tomoTestMaoz.py", line 203, in <module>
    proj_data = tomopy.remove_outlier3d(proj_data, dif=0.5, size=3)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Users\yakovlev\AppData\Local\anaconda3\envs\TomoRecon\Lib\site-packages\tomopy\misc\corr.py", line 458, in remove_outlier3d
    extern.c_median_filt3d_float32(np.ascontiguousarray(arr), out, kernel_half_size, dif, ncore,
  File "C:\Users\yakovlev\AppData\Local\anaconda3\envs\TomoRecon\Lib\site-packages\tomopy\util\extern\misc.py", line 125, in c_median_filt3d_float32
    LIB_TOMOPY_MISC.medianfilter_main_float(
OSError: exception: access violation reading 0x000002BCCEB9F020
```

**Expected behavior**
For remove_outlier3d to return a numpy array the same size as the input projection stack, with pixels deviating from the local containing 3-pixel median by 0.5 or more to be set to the median value.

**Helpful images**
Posted above

**Platform Information:**
 - OS: Windows 11
 - Python Version 3.11.5
 - TomoPy Version 1.14.1

**Additional context**
Issue was reproduced on python version 3.7

Thanks!

dkazanc:
The main issue was around `memcpy` seg faulting, possibly when the total index size overflows `int`. 

But in fact, there were multiple issues:
- as previously the index could overflow I introduced `size_t` data type, which should work well across different OS.
- In the wrapper there was `np.ascontiguousarray(arr), out` but should be `np.ascontiguousarray(arr), np.ascontiguousarray(out)`.
- Removed memcpy in the C code and copied array directly in the Python wrapper.

The combination of above seemed solved the problem.

Fixes #622
Closes #624

### DIFF
--- a/source/libtomo/misc/median_filt3d.c
+++ b/source/libtomo/misc/median_filt3d.c
@@ -66,7 +66,7 @@ int uint16comp(const void* elem1, const void* elem2)
 
 void
 medfilt3D_float(float* Input, float* Output, int radius, int sizefilter_total,
-                float mu_threshold, long i, long j, long k, long index, long dimX,
+                float mu_threshold, long i, long j, long k, size_t index, long dimX,
                 long dimY, long dimZ)
 {
     float*    ValVec;
@@ -78,6 +78,7 @@ medfilt3D_float(float* Input, float* Output, int radius, int sizefilter_total,
     long      k1;
     long long counter;
     int       midval;
+    size_t    index1;
     midval = (sizefilter_total / 2);
     ValVec = (float*) calloc(sizefilter_total, sizeof(float));
 
@@ -98,7 +99,8 @@ medfilt3D_float(float* Input, float* Output, int radius, int sizefilter_total,
                 k1 = k + k_m;
                 if((k1 < 0) || (k1 >= dimZ))
                     k1 = k;
-                ValVec[counter] = Input[(dimX * dimY) * k1 + j1 * dimX + i1];
+                index1 = (size_t)(dimX * dimY * k1 ) + (size_t)(j1 * dimX + i1);
+                ValVec[counter] = Input[index1];
                 counter++;
             }
         }
@@ -122,7 +124,7 @@ medfilt3D_float(float* Input, float* Output, int radius, int sizefilter_total,
 
 void
 medfilt2D_float(float* Input, float* Output, int radius, int sizefilter_total,
-                float mu_threshold, long i, long j, long index, long dimX, long dimY)
+                float mu_threshold, long i, long j, size_t index, long dimX, long dimY)
 {
     float*    ValVec;
     long      i_m;
@@ -131,6 +133,7 @@ medfilt2D_float(float* Input, float* Output, int radius, int sizefilter_total,
     long      j1;
     long long counter;
     int       midval;
+    size_t    index1;
     midval = (sizefilter_total / 2);
     ValVec = (float*) calloc(sizefilter_total, sizeof(float));
 
@@ -146,7 +149,8 @@ medfilt2D_float(float* Input, float* Output, int radius, int sizefilter_total,
             j1 = j + j_m;
             if((j1 < 0) || (j1 >= dimY))
                 j1 = j;
-            ValVec[counter] = Input[j1 * dimX + i1];
+            index1 = (size_t)(j1 * dimX + i1);
+            ValVec[counter] = Input[index1];
             counter++;
         }
     }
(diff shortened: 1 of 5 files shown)


source/tomopy/misc/corr.py
```diff
@@ -441,8 +442,8 @@ def remove_outlier3d(arr, dif, size=3, ncore=None):
     input_type = arr.dtype
     if (input_type != 'float32') and (input_type != 'uint16'):
         arr = dtype.as_float32(arr)  # silent convertion to float32 data type
-    out = np.empty_like(arr)
-
+    out = copy.deepcopy(arr)
```
carterbox, line 445: Use `np.copy(arr, order='C')`, instead of making a copy then checking if the copy is contiguous. (If it's not, then you would end up copying the array twice!?)

source/libtomo/misc/median_filt3d.c
```diff
@@ -98,7 +100,8 @@ medfilt3D_float(float* Input, float* Output, int radius, int sizefilter_total,
                 k1 = k + k_m;
                 if((k1 < 0) || (k1 >= dimZ))
                     k1 = k;
-                ValVec[counter] = Input[(dimX * dimY) * k1 + j1 * dimX + i1];
+                index1 = (size_t)(dimX * dimY * k1 ) + (size_t)(j1 * dimX + i1);
```
carterbox, line 103: What's the pro/con of using `size_t` for all indexing and size variables instead of a mix of long and size_t with casting? Do some of these indexing variable need to represent negative integers?

My concern with mixing types and casting on lines like this is that it might still overflow before the cast to size_t? Because the order of operations is to perform the computation inside the parenthesis before casting.

Maybe it should be something like this? Where everything that is not cast is already `size_t`? This way the multiplications are computed using types that are already `size_t`.
```suggestion
                index1 = dimX * dimY * (size_t)k1 + (size_t)j1 * dimX + (size_t)i1;
```

carterbox:
Thanks for starting a fix for this @dkazanc! I was waiting for confirmation that the problem was reproducable with arbitrary arrays. Questions/comments above.

source/tomopy/misc/corr.py
dkazanc: thanks, yes you're right. I guess we could do something like this `np.ascontiguousarray(arr), np.copy(arr, order='C')` in the list of parameters

carterbox: No. We need the reference to the variable `out` so we can return it on line 463.

source/libtomo/misc/median_filt3d.c
dkazanc: thanks, makes total sense. I guess `(size_t)(dimX * dimY * k1 )` could overflow before getting converted to `size_t`. will try to fix...

dkazanc:
@carterbox are you OK with the corrections? thx

carterbox (approved):
Thanks, @dkazanc! Sorry for the broken CI stuff.

PR #637, opened 2024-02-05
REF: Replace interp2d deprecated

nghia-vo:
This implementation is closest in terms of performance and quality compared to interp2d.


carterbox:
Thanks, @nghia-vo

PR #640, opened 2024-04-12
DOC: added doi to gridrec papers

decarlof:


carterbox:
The scipy.signal API has broken between 1.9 and 1.13

Issue #650, opened 2024-10-25
AttributeError from deprecated np.bool alias with numpy>=1.24.0


fgjorup:
Numpy raises and error, `AttributeError: module 'numpy' has no attribute 'bool'` due to deprecated aliases of built-in type [numpy release note 1.20.0](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated)

**Code**
```python
import tomopy
import numpy as np

proj = np.random.random((180,1,180))
proj = tomopy.misc.morph.pad(proj, axis=2, mode='edge')

```
**Error message**
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[4], line 2
         1 proj = np.random.random((180,1,180))
----> 2 proj = tomopy.misc.morph.pad(proj, axis=2, mode='edge')

/python3.11/site-packages/tomopy/misc/morph.py:139, in pad(arr, axis, npad, mode, ncore, **kwargs)
      136 slc_r_v = tuple(slc_r_v)
      138 out = np.empty(newshape, dtype=arr.dtype)
--> 139 if arr.dtype in [np.float32, np.float64, np.bool,
      140                  np.int32, np.int64, np.complex128]:
      141     # Datatype supported by numexpr
      142     with mproc.set_numexpr_threads(ncore):
      143         ne.evaluate("arr", out=out[slc_in])

/python3.11/site-packages/numpy/__init__.py:324, in __getattr__(attr)
      319     warnings.warn(
      320         f"In the future `np.{attr}` will be defined as the "
      321         "corresponding NumPy scalar.", FutureWarning, stacklevel=2)
      323 if attr in __former_attrs__:
--> 324     raise AttributeError(__former_attrs__[attr])
      326 if attr == 'testing':
      327     import numpy.testing as testing

AttributeError: module 'numpy' has no attribute 'bool'.
`np.bool` was a deprecated alias for the builtin `bool`. To avoid this error in existing code, use `bool` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.bool_` here.
The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations
```

**Suggested solution**
_np.bool_ should the changed to _bool_ to be compatible with newer numpy versions.

**Platform Information:**
[...]

PR #651, opened 2024-11-13
Update morph.py

decarlof:
addressing #650


carterbox:
It looks like conda is no longer provided in the macOS images. It was dropped sometime after macOS-12 and the current latest is macOS-14.

https://github.com/actions/runner-images/tree/main

